PR #1, opened 2017-01-23
Enable deploy from Travis

annevk:
This moves the whole website into /src for easy of copying. And adds a
script to deploy from Travis when a commit lands in master.


deploy.sh
```diff
@@ -0,0 +1,13 @@
...
+ENCRYPTED_KEY_VAR="encrypted_${ENCRYPTION_LABEL}_key"
+ENCRYPTED_IV_VAR="encrypted_${ENCRYPTION_LABEL}_iv"
+ENCRYPTED_KEY=${!ENCRYPTED_KEY_VAR}
+ENCRYPTED_IV=${!ENCRYPTED_IV_VAR}
+openssl aes-256-cbc -K $ENCRYPTED_KEY -iv $ENCRYPTED_IV -in deploy_key.enc -out deploy_key -d
+chmod 600 deploy_key
+eval `ssh-agent -s`
+ssh-add deploy_key
+echo "$SERVER $SERVER_PUBLIC_KEY" > known_hosts
+rsync -avz -e "ssh -o UserKnownHostsFile=known_hosts" src/ $DEPLOY_USER@$SERVER:$WEB_ROOT/
```
domenic: Nit: expanding command line flags makes things easier for future readers. So maybe

--archive --verbose --compress --rsh="ssh ..."

annevk:
I don't understand. Travis seems to keep deploying this, even when the branch is not master, because it checks if it can be done on top of master.

However, since we use the same checks as for standards that would mean whatever the latest PR is would end up in a standard, which doesn't seem right.

@sideshowbarker @domenic?

domenic:
Seems to be working fine: https://travis-ci.org/whatwg/whatwg.org#L155

"Skipping deploy for a pull request"

annevk:
It shows two checks have passed. https://travis-ci.org/whatwg/whatwg.org/builds/194760632 skips. https://travis-ci.org/whatwg/whatwg.org/builds/194760618 ran rsync. Doesn't that mean something is wrong?

domenic:
Oh, I see. Strange. We'll definitely have to check the others.

Given that this is meant to only run on Travis, I'd suggest simplifying the check as follows:

if [ "$TRAVIS_BRANCH" != "master" -o "$TRAVIS_PULL_REQUEST" != "false" ]

You may also want to echo $TRAVIS_BRANCH and $TRAVIS_PULL_REQUEST for debugging.

annevk:
(I did check URL by the way and I couldn't reproduce what was happening here. It didn't generate the Living Standard copy.)

annevk:
Yeah, but the results are still bonkers for some reason.

annevk:
@domenic feel free to squash and merge or suggest further tweaks or let me squash and merge it tomorrow. I think it's all good finally. Silly typo.

PR #17, opened 2017-02-14
Replace reference to developers.whatwg.org with blog.whatwg.org

### LINKED ISSUE #9: Remove references to developers.whatwg.org for now
We should still work on getting the developers spec back and running, but until we do, we should not link to it.

annevk:
Fixes #9. Also reference checker.html5.org rather than validator.nu as
@sideshowbarker keeps that more up-to-date.

### DIFF
--- a/src/index
+++ b/src/index
@@ -3,20 +3,20 @@
  <head>
   <title>Web Hypertext Application Technology Working Group</title>
   <link rel="stylesheet" href="/style/tabbed-pages">
-  <link rel="icon" href="/images/icon">
+  <link rel="icon" href="https://resources.whatwg.org/logo.svg">
  </head>
  <body class="front-page">
   <hgroup>
-   <h1><object data="https://images.whatwg.org/logo.svg" height=101 width=101><img src="https://images.whatwg.org/logo" alt="" height=101 width=101></object> Welcome to the WHATWG community</h1>
+   <h1><img src="https://resources.whatwg.org/logo.svg" height=101 width=101 alt> Welcome to the WHATWG community</h1>
    <h2>Maintaining and evolving HTML since 2004</h2>
   </hgroup>
 
   <a href="https://wiki.whatwg.org/wiki/FAQ"><p><strong>FAQ</strong> <span>Get answers</span> <span>to your questions</span> <span>about HTML</span> <span>and the WHATWG</span></p></a>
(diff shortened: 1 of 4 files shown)


src/index
```diff
@@ -3,20 +3,20 @@
...
   <a href="https://html.spec.whatwg.org/multipage/"><p><strong>HTML</strong> <span>Read, use,</span> <span>or implement</span> <span>the HTML Living Standard</span></p></a>
-  <a href="https://whatwg.org/specs/"><p><strong>Specs</strong> <span>See the other</span> specifications <span>developed at the WHATWG</span></p></a>
+  <a href="https://whatwg.org/specs/"><p><strong>Standards</strong> <span>See the other</span> standards <span>developed at the WHATWG</span></p></a>
 
-  <a href="https://developers.whatwg.org/"><p><strong>Web Dev Edition</strong> <span>Read an edition of</span> <span>the HTML standard</span> <span>designed for</span> <span>Web Developers</span></p></a>
-  <a href="http://html5.validator.nu/"><p><strong>Validator</strong> <span>Check your documents</span> <span>with an HTML validator</span></p></a>
+  <a href="https://blog.whatwg.org/"><p><strong>News</strong> <span>Read and contribute</span> <span>to the WHATWG blog</span></p></a>
+  <a href="https://checker.html5.org/"><p><strong>Html Checker</strong> <span>Validate your documents</span> <span>with an HTML checker</span></p></a>
```
zcorpan: All-uppercase "HTML"?

annevk: I guess @sideshowbarker should help here, since the page doesn't spell it that way.

zcorpan: It also says "HTML checking", but yeah

sideshowbarker: Yeah the “Html” in the formal title of the checker service is intentional. But I don’t feel strongly that it needs to preserved that way in references to the service.

domenic: It seems really strange to me; I'd prefer "HTML". Why is it that way in the checker service, I wonder?

sideshowbarker: > Why is it that way in the checker service, I wonder?

Because it looks better in the full context: _Nu Html Checker_ and I like it better that way. But again I don’t feel strongly about forcing anybody else to replicate that faithfully in references.

domenic: Yeah I guess if the link here was titled "nu html checker" we could go with that casing but since it's just "html checker" as a generic term uppercase makes more sense.

annevk:
Addressed feedback I think.

PR #24, opened 2017-04-01
Typo fix

xfq:
`padding; 0;` -> `padding: 0;`.


annevk:
That whole line can be dropped since we no longer use object and it's very unlikely we ever would again in the future.

xfq:
Done. Thanks for the review!

annevk:
Thank you for your attention to detail!

xfq:
You're welcome.

PR #30, opened 2017-05-24
Add parse error table style

inikulin:
Required for https://github.com/whatwg/html/pull/2701#discussion_r117782153

### DIFF
--- a/src/style/specification
+++ b/src/style/specification
@@ -3,7 +3,7 @@
 html { margin: 0; padding: 0; background: white; color: black; }
 body { margin: 0 auto; padding: 0 1.5em 2em 2.5em; max-width: 80em; background: white; color: black; }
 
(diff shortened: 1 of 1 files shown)


inikulin:
Hmm, seems like file had mixed line endings. My editor changed them to LF.

domenic:
Since this is HTML-specific, I think it would make sense in the `<style>` block at the top of the HTML source file?

annevk:
Also, is HTML not using standard.css from https://github.com/whatwg/resources.whatwg.org? If not, we should maybe fix that.

inikulin:
> Since this is HTML-specific, I think it would make sense in the <style> block at the top of the HTML source file?

Oh, I've missed an elephant in the room. Will move it there then. Closing this one.

PR #31, opened 2017-05-29
Move validator.whatwg.org here and update it

zcorpan:


hsivonen:
It seems a bit weird to remove Validator.nu from the list of "known" validators if the list is portrayed as the list of known ones. If you wish to point people to only one instance of that codebase, I suggest framing the single suggestion differently.

zcorpan:
Yes, you are right. We could list both, and maybe even list the W3C-hosted instance, and explain how they are different. Or list one that we think is generally most up to date and tracks whatwg/html rather than w3c/html.

domenic:
I don't think we should suggest anything that doesn't track the HTML Standard.

Otherwise, I'm agnostic on whether we should list more than one shell around a single codebase. My impression was validator.nu was not as up-to-date as checker.html5.org, but I don't remember where I heard that, so maybe it's not accurate.

sideshowbarker:
While it’s true  https://checker.html5.org/ is updated more often I guess it could be characterized as tracking the spec a bit over-aggressively—to the point of sometimes causing problems and confusion for users, because, e.g., I occasionally inadvertently introduce some significant regressions.

So https://validator.nu/ doesn’t get updated quite as often, that’s not completely a bad thing—because it’s more reliable as far as not having the kind of occasional temporary-though-serious regressions that https://checker.html5.org/ sometimes does (and modulo some problems with  https://validator.nu/ getting wedged now and then, and needing to ping @hsivonen to restart it).

zcorpan:
OK, see #33.

PR #36, opened 2017-08-09
Move the FAQ here

foolip:
This was produced starting from the HTML output of
https://wiki.whatwg.org/wiki/FAQ and cleaning it up to match the style
of /working-mode and /code-of-conduct. Only very minor visible changes
were made around whitespace, quotes and similar. Actual changes to the
FAQ will be made separately.

Some minimal amount of style is carried over or invented to make it not
much worse than the old FAQ.

All IDs are preserved.

### DIFF
--- a/src/.htaccess
+++ b/src/.htaccess
@@ -84,7 +84,6 @@ redirect permanent /x https://xhr.spec.whatwg.org/
 redirect permanent /xhr https://xhr.spec.whatwg.org/
 
 redirect permanent /newbug https://github.com/whatwg/html/issues/new
-redirect permanent /faq https://wiki.whatwg.org/wiki/FAQ
 
 # redirect permanent /) https://whatwg.org/
 redirect permanent /mailing-list) https://whatwg.org/mailing-list
--- a/src/code-of-conduct
+++ b/src/code-of-conduct
@@ -16,6 +16,7 @@
    <li><a href="https://spec.whatwg.org/">Standards</a></li>
    <li><a href="/working-mode">Working Mode</a></li>
    <li class="this"><strong>Code of Conduct</strong></li>
+   <li><a href="/faq">FAQ</a></li>
   </ul>
 
   <h2>Code of Conduct</h2>
--- a/src/faq
+++ b/src/faq
@@ -1,0 +1,315 @@
+<!doctype html>
+<html lang="en">
+ <head>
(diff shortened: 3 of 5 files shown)


foolip:
Like with some Bikeshed conversions, I compared the textual before/after and IDs using various scripts, so I'm pretty confident I didn't mess anything up badly.

foolip:
Pushed another change so that there are two commits now. They shouldn't be squashed when merging, use rebase.

src/.htaccess
```diff
@@ -84,7 +84,6 @@ redirect permanent /x https://xhr.spec.whatwg.org/
 redirect permanent /xhr https://xhr.spec.whatwg.org/
 
 redirect permanent /newbug https://github.com/whatwg/html/issues/new
-redirect permanent /faq https://wiki.whatwg.org/wiki/FAQ
```
domenic: Uh oh. Given cached redirects, we may need a new URL...

domenic:
Other thoughts: we may want to add self-links, like we did in working-mode. And we may want to mint new IDs, as following this wiki-generated convention for future edits is going to be painful.

src/.htaccess
annevk: User agents tend not to cache them long so I think we should be fine. We could go with frequently-asked-questions though if we wanted to. Would be consistent with code-of-conduct and working-mode.

foolip:
I'm going to remove the mailing list part of this as well, instructions for Outlook aren't that helpful. I'll leave it on the wiki when updating it to point to the new FAQs.

src/.htaccess
foolip: I was hoping that /faq it was rarely linked, but https://html.spec.whatwg.org/ does. But, I'd still like to use the shorter URL, any breakage will be transient.

foolip:
This is now ready for review again, I've minted new IDs and done some more cleanup. Still no real changes to the actual content of course.

src/.htaccess
domenic: Shorter URL is nice. If you all aren't worried then I guess I am not either.

foolip:
I enjoy Git surgery, will try to turn it into something nicer and merge tomorrow.

foolip:
OK, I went with two commit, because once the HTML parts were out of the FAQ, the resulting second commit wasn't too big.

PR #47, opened 2017-08-22
Get rid of a number of mailing lists

annevk:


annevk:
@Hixie what you do you think? Disable help@, implementors@, and commit-watchers@, as they're no longer used much or obsolete. For help@ which still sees some traffic we suggest to use Stack Overflow instead.

src/mailing-list
```diff
@@ -39,34 +33,7 @@
...
-
-   <h3>Unsubscribing</h3>
-
-   <p>To unsubscribe from the mailing list, send an e-mail
-   to <a href="mailto:help-unsubscribe@whatwg.org">help-unsubscribe@whatwg.org</a>
-   from the address you want to unsubscribe, and then reply to the
-   confirmation e-mail you receive.
-
-  </div>
+   then you want to use <a href="https://stackoverflow.com/">Stack Overflow</a>.</p>
```
foolip: s/you want to use/please use/ would sound a bit nicer I think.

annevk:
Sounds good, note that @Hixie needs to sign this off as he can actually disable the mailing lists.

domenic:
Not necessarily in this PR, but I realized that the remaining content of https://wiki.whatwg.org/wiki/FAQ could probably go here.

Hixie:
I suggest we e-mail the help@ list first to get people on board, see if they're ok with moving to stack overflow or whatever.

The other two lists aren't used currently so it should be fine (though we should probably send one last e-mail to the implementors@ list just as a heads-up to anyone still on it).

annevk:
Emails sent.

ldexterldesign:
Hey all,

Hope you're well

Apologies, slightly off topic

I'm streamlining my mailbox and currently subscribed to a number of whatwg mailing lists

Are mailing lists still where the primary whatwg conversation is happening or are there alternatives I can take a look at?

Hope to hear back

Cheers

annevk:
Hi @ldexterldesign, mailing lists are in fact not where most of the conversion takes place. The conversation takes place in GitHub repositories corresponding to each of the [standards](https://spec.whatwg.org/) we publish. You can find the link to the GitHub repository at the top of the standard. You can also browse all the GitHub repositories at https://github.com/whatwg.

There's no difference between the @lists.whatwg.org and @whatwg.org addresses.

ldexterldesign:
Hey @annevk,

Thanks for reply

Okay, I'll reconsider my mailing list subscriptions

> There's no difference between the @lists.whatwg.org and @whatwg.org addresses.

Are you sure, because my email suggests otherwise?:

screen shot 2017-09-21 at 06 27 11
screen shot 2017-09-21 at 06 27 14

Hope to hear back

Cheers

annevk:
I don't know. That may just be how Thunderbird classifies things or maybe DreamHost changed the configuration at some point.

ldexterldesign:
I filter emails into those folders based on the recipient:

Emails to whatwg@listswhatwg:

screen shot 2017-09-21 at 10 42 41

Emails to whatwg@whatwg:

screen shot 2017-09-21 at 10 42 48

I would suggest it's the latter of your guess, but it would be good to know for sure

Cheers

ldexterldesign:
BUMP

annevk:
I don't think anyone is really in a position to figure this out. At some point we'll move the mailing list away from DreamHost and we can make this right.

PR #49, opened 2017-08-22
Handle 410s

annevk:


src/404
```diff
@@ -1,7 +1,7 @@
 <!doctype html>
 <html lang="en">
  <head>
-  <title>Web Hypertext Application Technology Working Group: File Not Found</title>
+  <title>Web Hypertext Application Technology Working Group File Not Found</title>
```
domenic: Why remove the :?

annevk: Because we don't use that anywhere else.

PR #54, opened 2017-08-23
Improve the FAQ

annevk:
In particular:

* Point out more clearly we're about standards and tests.
* Point to Working Mode.
* Stop pointing to the mailing list.
* Stop using the word "spec".

### DIFF
--- a/src/faq
+++ b/src/faq
@@ -228,7 +228,7 @@
   added to them over time, at a rate intended to keep the standard a little ahead of the
   implementations, but not so far ahead that the implementations give up.
 
-  <p>Despite the continuous maintenance, or maybe we should say <i>as part</i> of the continuing
+  <p>Despite the continuous maintenance, or maybe we should say <em>as part</em> of the continuing
   maintenance, a significant effort is placed on getting the standard and the implementations to
   converge — the parts of the standard that are mature and stable are not changed willy nilly.
   Maintenance means that the days where the standard are brought down from the mountain and remain
--- a/src/faq
+++ b/src/faq
@@ -33,15 +33,14 @@
 
   <h4 id="what-is-the-whatwg">What is the WHATWG?<a class="self-link" href="#what-is-the-whatwg"></a></h4>
 
-  <p>The Web Hypertext Application Technology Working Group (WHATWG) is a growing community of
-  people interested in evolving the Web. It focuses primarily on the development of HTML and APIs
-  needed for Web applications.
+  <p>The Web Hypertext Application Technology Working Group (WHATWG) is a community of people
+  interested in evolving the web through standards and tests.
 
   <p>The WHATWG was founded by individuals of Apple, the Mozilla Foundation, and Opera Software in
   2004, after a W3C workshop. Apple, Mozilla and Opera were becoming increasingly concerned about
-  the W3C’s direction with XHTML, lack of interest in HTML and apparent disregard for the needs of
-  real-world authors. So, in response, these organisations set out with a mission to address these
-  concerns and the Web Hypertext Application Technology Working Group was born.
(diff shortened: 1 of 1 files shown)


src/faq
```diff
@@ -204,35 +176,34 @@
...
+   suitable for implementors or better hooks in to the rest of the standard, etc.
 
    <li><p>Ask browser vendors to implement the newly specified solution, even if it’s just an
    experimental implementation. This implementation experience usually means that new problems are
    found with the solution that need to be addressed, or that a different solution is actually
    better.
 
-   <li><p>Write a test suite for the feature to see if the implementations match the spec. This
-   usually highlights bugs in the implementations and also bugs in the spec.
+   <li><p>Write a test suite for the feature to see if the implementations match the standard. This
```
zcorpan: Move up this step?

src/faq
```diff
@@ -85,41 +84,14 @@
 
   <h4 id="how-does-the-whatwg-work">How does the WHATWG work?<a class="self-link" href="#how-does-the-whatwg-work"></a></h4>
 
-  <p>People <a href="https://github.com/whatwg">collaborate on GitHub</a> or send email to
-  <a href="/mailing-list#specs">the mailing list</a>.
-
-  <p>Each standard has one or more editors, who are responsible for dealing with feedback for that
```
domenic: I think some useful content was lost here, especially with the more storytelling style instead of the list-of-requirements style of the working mode document. It might be better to just remove it as it's gotten outdated, but maybe something like this could be added after the working mode:

> In brief, each standard has one or more [editors](https://whatwg.org/working-mode#editor), who guide its evolution. Like a software project, a WHATWG standard undergoes [continual maintenance](https://whatwg.org/faq#living-standard), and occasionally gets [new features](https://whatwg.org/working-mode#additions). This work is driven by the community and in collaboration with implementers.

I'm OK leaving it off though, or modifying the above.

src/faq
```diff
@@ -244,36 +215,36 @@
 
   <h4 id="living-standard">What does “Living Standard” mean?<a class="self-link" href="#living-standard"></a></h4>
 
-  <p>The WHATWG specifications are described as Living Standards. This means that they are standards
-  that are continuously updated as they receive feedback, either from Web designers, browser
+  <p>The WHATWG standards are described as Living Standards. This means that they are standards
+  that are continuously updated as they receive feedback, either from web designers, browser
```
domenic: s/designers/developers while here?

src/faq
```diff
@@ -244,36 +215,36 @@
...
   vendors, tool vendors, or indeed any other interested party. It also means that new features get
-  added to them over time, at a rate intended to keep the specifications a little ahead of the
-  implementations but not so far ahead that the implementations give up.
+  added to them over time, at a rate intended to keep the standard a little ahead of the
+  implementations, but not so far ahead that the implementations give up.
 
   <p>Despite the continuous maintenance, or maybe we should say <i>as part</i> of the continuing
```
domenic: i -> em while here?

annevk: I think it matches the definition of i more than em.

src/faq
annevk: I figured we care about both. I could add "and developers"?

src/faq
annevk: I'd be happy to include this.

src/faq
domenic: I suppose, although I was thinking of our general move from authors to developers.

src/faq
domenic: I don't want to spend any time debating this, so please disregard my feedback here for merging purposes. But I am idly curious what part of the definition of i you think this matches.

annevk: I was thinking alternate voice, but I'm happy to go with `<em>` now.

PR #55, opened 2017-08-23
Remove .htaccess for images.whatwg.org and n.whatwg.org

foolip:
The config will be elsewhere:
https://github.com/whatwg/misc-server/commit/c61552ac840b1fb99031a9252945da287cf66180

### DIFF
--- a/images.whatwg.org/.htaccess
+++ b/images.whatwg.org/.htaccess
@@ -1,2 +1,0 @@
-# Indexing Options
-IndexOptions DescriptionWidth=* NameWidth=* FancyIndexing FoldersFirst ScanHTMLTitles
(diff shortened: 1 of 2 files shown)


annevk:
So we might need to create some guidelines in the README about this kind of thing by the way. It won't matter for these legacy domains, but once we start spreading this we should have guidance about what goes where.

foolip:
My desire is to have no .htaccess files at all, if we can get there in a resonably short time there wouldn't be any documentation required, .htaccess simply wouldn't work. Although I suppose we should point to the misc-server repo from some places that people might look.

annevk:
It'll be a bit of work to get there for whatwg.org.

We should probably merge misc-server in here in due course.

domenic:
No .htaccess would be great as it'd help if we ever move to nginx. (Which has a single config file per server IIUC.)

I think HTML and maybe Encoding have subdirectory .htaccess files right now.

foolip:
> We should probably merge misc-server in here in due course

Yeah, maybe. In a config/ directory or something?

annevk:
Yeah, something like that, no need to figure it out now.

PR #56, opened 2017-08-24
Deploy to /var/www using a deploy user

foolip:


foolip:
chmod is needed to make sure it's world readable when it arrives, because the user running the server (www-data) otherwise can't read it. Not sure why it worked with the admin user, maybe this umask thing that I've never understood.

PR #60, opened 2017-08-25
Extract out spec.whatwg.org to its own top-level dir

foolip:
A copy is made to keep this working on both the old and new server.


annevk:
You mean whatwg.org/specs/ redirects?

We can't land this until spec.whatwg is on the new server though.

foolip:
> You mean whatwg.org/specs/ redirects?

Oops, update the commit message. (I copied URL from by browser, after redirect :)

> We can't land this until spec.whatwg is on the new server though.

It'll still deploy to the old server. The comment about "Removing whatwg.org/specs/index is OK" means to say that it's OK that the index file will disappear from the old server, because it won't be served except by https://whatwg.org/specs/index

annevk:
No it won't deploy to the older server. It'll get removed due to rsync. The way spec.whatwg.org is setup now is that it serves the file from whatwg.org/specs/. So if you move that file elsewhere, spec.whatwg.org will break.

foolip:
Oh, you're right. Let me try again.

foolip:
Done.

PR #61, opened 2017-08-26
Add .editorconfig and .gitattributes

### LINKED ISSUE #18: Add .editorconfig and .gitattributes to this repo

xfq:
Mostly stolen from the Streams repo.

Fix #18.


annevk:
Thanks, I'll let @domenic confirm since he's a little more familiar with these.

.editorconfig
```diff
@@ -0,0 +1,16 @@
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+charset = utf-8
+indent_size = 1
```
zouhir: this means all JavaScript, CSS & HTML, etc.. indentation will be 1 space only which is bit unusual maybe we should ask @domenic in #18 about his preference.

xfq: Yeah. Currently, most (if not all) HTML files use 1 space only; JavaScript files and inline `<script>` elements use 2 spaces; `.htaccess` files are not indented consistently. And I don't think we need to worry about the indentation of SVG files in this case.

I'll wait for Domenic's reply and adjust the PR if necessary.

domenic: 1 space is preferred for HTML (or SVG). I guess to be consistent we should also do it for CSS and script and .htaccess, especially since we don't use file extensions so differentiating would be tricky (see #8).

I think we should fix those files to be 1 space while merging this if you're up for it. Otherwise I think adding .editorconfig will just cause confusion since it will cause peoples' text editors to start being inconsistent with the indentation inside a file.

zcorpan: Can we have two spaces for .js ?

xfq: @zcorpan Done.

domenic:
My main concern is not having .editorconfig contradict the indentation we want. I.e. if we ever edit those files, we should not be fighting our editor which insists on 1 or 2 spaces per .editorconfig.

I think the best solution for that is to consistify things. But an alternate solution would be to add a lot of exclude rules. Or switch to a safelist approach where we only add indent settings for certain known-good files.

annevk:
Closing due to inactivity. This would still be welcome, but probably needs to be more tightly scoped.

PR #62, opened 2017-08-27
Merge resources.whatwg.org repo into a subdirectory of whatwg.org

foolip:
This merges https://github.com/whatwg/resources.whatwg.org/commit/37eac78519af647f70a98149512bc03a2216ba98 into this repo under resources.whatwg.org.


foolip:
This creates a merge commit, and there were none in this repo before. If preferred, I can create a single commit pointing back to what it was copied from.

foolip:
(Deploy changes aren't part of this, some vetting of the differences between what's in git and on the server is needed.)

foolip:
I went ahead and added deploy but without `--delete`. Before this is merged, the webhook for resources.whatwg.org should be disabled.

annevk:
Since we want resources.whatwg.org to remain static I think this makes sense. I guess we'd delete the repository after this and potentially break some links? Or would we just mark it obsolete?

foolip:
I think we should do a final commit that points to this repo. And migrate the remaining issues I suppose.

resources.whatwg.org/.htaccess
```diff
@@ -0,0 +1,5 @@
+Options Indexes FollowSymLinks ExecCGI
+IndexIgnore booyah.cgi
```
annevk: This line can be removed.

PR #64, opened 2017-08-28
Add a tracking vector image

zcorpan:
Fixes https://github.com/whatwg/resources.whatwg.org/issues/49

---

This is taken from https://github.com/whatwg/resources.whatwg.org/pull/50. It's blocked on https://github.com/tabatkins/bikeshed/issues/964 (or at least I wanted to merge all fingerprint-related PRs at the same time).


annevk:
Should we also redirect images.whatwg.org/fingerprint.png to this?

foolip:
There's a very large overlap between https://images.whatwg.org/ and https://html.spec.whatwg.org/images/, although https://github.com/whatwg/html/commit/3cc2ca695e052daa0681e36439a1c9f634a78b56 only happened on one side so there are some differences.

Do we want to use images.whatwg.org for anything going forward, or should it all be in either the spec's repo or resources.whatwg.org?

annevk:
images is largely historical. I suspect a lot of old specifications hosted on whatwg.org still use it.

annevk:
To reply to the very old comments: I don't think we should do anything here with images.whatwg.org. Let's leave it alone.

I suggest we land the image @zcorpan created as `tracking.svg` and also rename the classes as appropriate and standardize around that. I'm happy to take on that work.

zcorpan:
LGTM

snuggs:
LGTM. Far less semantic/understandable until @annevk changed the name. /cc @sideshowbarker

annevk:
I'm going to merge this as it doesn't affect the HTML Standard as I thought it would. That currently has inline CSS for fingerprint.

Issue #65, opened 2017-08-28
Fetching deploy.sh keeps failing


annevk:
This is for Fetch's master:
```
Setting environment variables from .travis.yml
$ export ENCRYPTION_LABEL="e149beb9c312"
$ export DEPLOY_USER="annevankesteren"

$ bash -c 'echo $BASH_VERSION'
4.3.11(1)-release

0.43s$ curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh && bash ./deploy.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  5503  100  5503    0     0  21895      0 --:--:-- --:--:-- --:--:-- 21924
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100    87  100    87    0     0    574      0 --:--:-- --:--:-- --:--:--   576

The command "curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh && bash ./deploy.sh" exited with 1.

Done. Your build exited with 1.
```
Fetch it from `https://raw.githubusercontent.com/whatwg/whatwg.org/master/resources.whatwg.org/build/deploy.sh` instead? And if we go down that path, no longer host it on `resources.whatwg.org` as we fetch it from GitHub anyway?

cc @domenic @foolip

annevk:
https://travis-ci.org/whatwg/fetch/builds/269117879

foolip:
Is this a flaky failure? I can currently access https://resources.whatwg.org/build/deploy.sh just fine.

foolip:
I've restarted the build to see.

annevk:
FWIW, I already had restarted it four times while being able to access that resource just fine myself. It's rather unclear what's going on.

foolip:
Might be because of the new deploy method for resources.whatwg.org, perhaps?

foolip:
But I can't exactly how that could be, since the resource is there.

annevk:
Nah, this happened before this switch as well.

foolip:
https://github.com/whatwg/fetch/pull/587 didn't give much of a clue :/

foolip:
Has this been happening for all repos?

annevk:
Yeah, they just haven't been updated much recently so I haven't hit it a lot.

annevk:
[...]

PR #67, opened 2017-08-28
Update link to standard-service-worker.js after repo merge

foolip:
See https://github.com/whatwg/whatwg.org/issues/65#issuecomment-325331716.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -49,7 +49,7 @@ if [[ "$TRAVIS" == "true" ]]; then # For some reason the above does not work on 
     BRANCH=$TRAVIS_BRANCH
 fi
 
-SERVICE_WORKER_SHA=$(curl https://api.github.com/repos/whatwg/resources.whatwg.org/contents/standard-service-worker.js \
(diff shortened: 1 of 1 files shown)


resources.whatwg.org/build/deploy.sh
```diff
@@ -49,7 +49,7 @@ if [[ "$TRAVIS" == "true" ]]; then # For some reason the above does not work on
     BRANCH=$TRAVIS_BRANCH
 fi
 
-SERVICE_WORKER_SHA=$(curl https://api.github.com/repos/whatwg/resources.whatwg.org/contents/standard-service-worker.js \
+SERVICE_WORKER_SHA=$(curl https://api.github.com/repos/whatwg/whatwg.org/contents/resources.whatwg.org/standard-service-worker.js \
```
foolip: This'll fix the problem, but `git hash-object resources.whatwg.org/standard-service-worker.js` would give the same result. @domenic, can I use that? Or just shasum? (The git sha also involves the size of the object, so not the same.)

foolip: Ha, I'm confused, this script isn't run in the same repo :)

PR #68, opened 2017-08-28
Support deploying extra files using an EXTRA_FILES variable

foolip:


resources.whatwg.org/build/deploy.sh
```diff
@@ -63,6 +63,12 @@ echo ""
 
 rm -rf "$WEB_ROOT" || exit 0
 
+copy_extra_files() {
+    if [[ "$EXTRA_FILES" != "" ]]; then
+        cp $EXTRA_FILES $1
```
domenic: It appears we're not running shellcheck on this script. (It does not like the grep regexp also.)

Per https://github.com/koalaman/shellcheck/wiki/SC2086, it suggests we quote at least the `$1`. It also wants us to quote the `$EXTRA_FILES`, but that defeats the point; I've asked about what the right pattern is at https://stackoverflow.com/q/45931553/3191 .

In the meantime, let's just quote `$1` at least.

domenic:
Also not sure what the plan is for https://github.com/whatwg/url/blob/master/.htaccess which should only be deployed to master, not the snapshots.

resources.whatwg.org/build/deploy.sh
foolip: Ah, thanks, wonder what the answer will be :)

foolip:
I didn't notice .htaccess was only for master, but happily .htaccess will be to move that out of the repo entirely to misc-server's configs, so this won't be a problem.

resources.whatwg.org/build/deploy.sh
```diff
@@ -25,6 +25,7 @@ NEW_SERVER_PUBLIC_KEY="ecdsa-sha2-nistp256 AAAAE2VjZHNhLXNoYTItbmlzdHAyNTYAAAAIb
 TRAVIS=${TRAVIS:-false}
 TRAVIS_PULL_REQUEST=${TRAVIS_PULL_REQUEST:-false}
 DEPLOY_USER=${DEPLOY_USER:-}
+EXTRA_FILES=${EXTRA_FILES:-}
```
foolip: Added this after local testing revealed an error: "EXTRA_FILES: unbound variable"

With that fix, it works locally, so merging this.

resources.whatwg.org/build/deploy.sh
foolip: Also adding a link to the question in the source.

PR #72, opened 2017-08-30
git adeploy.sh: move first logging to earlier to help debugging

foolip:
In a case like https://github.com/whatwg/compat/pull/82 it's not easy
to tell how far into deploy.sh we got, this should help.


annevk:
This seems fine to me, but @domenic should review since he made this all work.

Issue #73, opened 2017-08-31
Unauthenticated use of api.github.com is rate limited; fails on Travis


foolip:
(moved from the common-build repo that I'm going to delete)

As part of https://github.com/whatwg/whatwg.org/issues/71 I've seen some branch builds fail mysteriously, and after adding debugging it looks like it was failing in:
```sh
SERVICE_WORKER_SHA=$(curl https://api.github.com/repos/whatwg/whatwg.org/contents/resources.whatwg.org/standard-service-worker.js \
                     -H "Accept: application/vnd.github.v3+json" \
                     | grep -Po '(?<="sha": ")[^"]*') # Hacky JSON parsing but works for SHAs
```

It doesn't always reproduce, but in https://travis-ci.org/whatwg/infra/builds/270176543 I had logging and caught "curl: (22) The requested URL returned error: 403 Forbidden"

Most likely, this is due to rate limiting.

@domenic, before I attempt a fix, can you help me understand the purpose of this service worker freshness check? Is the purpose merely to change the generated service-worker.js in some (any) way if https://resources.whatwg.org/standard-service-worker.js has changed? Why does this make sense to do at build time, given that the standard could be updated much less frequently than standard-service-worker.js?

foolip:
https://developer.github.com/v3/ is why I suspect rate limits.

@domenic said, re the sha:

> That is indeed the purpose. As for why at build time, I'm not sure when else we would build the service worker?

foolip:
@domenic, about this sha, is the only thing that would cause the service worker script to be reloaded a change to /service-workers.js, and does that mean that for standards that very rarely change, they won't get any improvements to the service worker script?

If the script has to be hosted on the same origin, and changing its content (as opposed to headers) is the only way, then we could perhaps set up a daily travis cron job that touches these files?

domenic:
> does that mean that for standards that very rarely change, they won't get any improvements to the service worker script?

I suppose so. However if the spec itself hasn't changed I'm not sure it really needs better service worker behavior.

I was more worried about cases like, us introducing some new image into the template, and then updating the service worker to cache that. But if we don't rebuild the spec at all, then it would not get the new template, so it doesn't need to cache the new image.

> we could perhaps set up a daily travis cron job that touches these files?

Well, I don't want to cause unnecessary re-fetching either. Re-fetching the service worker code every day seems kind of sad.

PR #74, opened 2017-08-31
Simplify and delay standard-service-worker.js sha calculation

foolip:
Fixes https://github.com/whatwg/whatwg.org/issues/73.


foolip:
This was moved over from the common-build repo that I'm going to delete, was approved by @domenic there, so I'll merge this if Travis is happy.

foolip:
Ah, review is required, can't merge.

PR #77, opened 2017-08-31
Consolidate most specs/ .htaccess resources

annevk:
specs/web-forms/ is still remaining as it's much more involved.


whatwg.org/specs/.htaccess
```diff
@@ -1,5 +0,0 @@
-Redirect permanent /specs/html5 https://html.spec.whatwg.org/multipage/
-Redirect permanent /specs/vocabs https://html.spec.whatwg.org/multipage/microdata.html
-
-# polish version got way out of date
-redirect gone /specs/index.pl
```
annevk: Note that this is already broken as https://spec.whatwg.org/index.pl returns a 404. We could restore it there, but then @foolip would have to fill me in how to do that.

PR #82, opened 2017-09-01
deploy.sh: make the deploy server and public key configurable

foolip:
This is necessary to transition the console and streams spec, which are
not on 75.119.197.251.


foolip:
I've discovered that this actually isn't needed, I mistakenly assumed that the annevankesteren and ddenicola account were on different machines, but they aren't. Should I revert it to not have any more complexity than necessary?

domenic:
Either revert, or wait until we're completely moved to then new server and this all becomes unnecessary, I guess.

foolip:
I'll leave it in until the NEW_THINGS become the new THINGS, then.

PR #81, opened 2017-09-01
deploy.sh: support running a post build step

foolip:
Usage example for encoding, in .travis.yml:
```
 - POST_BUILD_STEP="python visualize.py \"$DIR/\""
```

Part of https://github.com/whatwg/meta/issues/11.

PR #86, opened 2017-09-03
deploy.sh: document optional environment variables like EXTRA_FILES

foolip:
Documentation for:
 * https://github.com/whatwg/whatwg.org/pull/68
 * https://github.com/whatwg/whatwg.org/pull/81
 * https://github.com/whatwg/whatwg.org/pull/82


foolip:
The post build step will be invoked 3 times, so to avoid duplicate work one should pass in the directory. We could arrange things such that the directory is always passed in after the other bits, but for encoding a trailing slash is needed, and I imagine things like `cp -a "$DIR" tmp; ...` where it's not a trailing argument, or perhaps is used more than once.

foolip:
Maybe https://github.com/whatwg/whatwg.org/pull/82 should be reverted, in which case part of this should be removed too. Opinions?

foolip:
Decided to keep these bits until they're removed again once the transition is complete. I'll merge this, as I want to change EXTRA_FILES and update the documentation that goes with it.

PR #92, opened 2017-09-16
Update URLs of Books and Figures in biblio.json

foolip:
See https://github.com/tobie/specref/issues/382.


foolip:
Oh, I didn't know this was a common format of some sort and that extra fields would do anything.

PR #94, opened 2017-09-18
Working Mode: encourage patience and due diligence

annevk:
As per https://github.com/whatwg/meta/issues/38.


domenic:
I am concerned with this phrasing being used to say "you can't close the issue, I don't agree yet". Maybe it is best to just leave this off.

annevk:
Okay, I guess we can leave it until it becomes a problem somehow.

PR #95, opened 2017-09-26
Fix PNG logo colors

domenic:
Some of the previous conversions were done with some sort of color setting which changed the output colors. This fixes them, and also adds a PNG version for compat.

Discovered in https://github.com/whatwg/meta/issues/28#issuecomment-332218130.


annevk:
I cannot see the difference. Is this some color profile thing?

domenic:
Yes, it must be. On my screen on Firefox and Chrome the "before" is (86, 118, 6) whereas the "after" is (60, 121, 10) i.e. WHATWG green.

annevk:
I'm not sure this really fixes it. On macOS I don't see a difference between before/after across browsers, but I do see (using Digital Color Meter):

Firefox: 60, 121, 10
Chrome/Safari: 75, 119, 36

domenic:
Hmm, I don't have macOS available, so I can't really help with that. Maybe we can merge this for now to improve Windows users, and try to fix macOS later?

PR #98, opened 2017-10-02
biblio.json: mark LOADER as an idea

annevk:


xfq:
It should also be added to `idea.whatwg.org/index.html`, I guess?

foolip:
Yes, but it'd be nice to do it after it's hosted on loader.idea.whatwg.org.

PR #105, opened 2017-10-22
Simplify .htaccess IndexIgnore/DirectoryIndex

foolip:
What remains is all that's still relevant, no files of the other patterns exist.

### DIFF
--- a/whatwg.org/.htaccess
+++ b/whatwg.org/.htaccess
@@ -14,8 +14,8 @@ AddType image/gif .gif
 # Indexing Options
 IndexOptions DescriptionWidth=* NameWidth=* FancyIndexing FoldersFirst ScanHTMLTitles
 # also add SupressHTMLPreamble once we use HeaderName and some fancy CGI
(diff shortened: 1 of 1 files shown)


foolip:
I'm converting these to nginx, and thought I'd make it clear that these did nothing in the first place.

whatwg.org/.htaccess
```diff
@@ -14,8 +14,8 @@ AddType image/gif .gif
 # Indexing Options
 IndexOptions DescriptionWidth=* NameWidth=* FancyIndexing FoldersFirst ScanHTMLTitles
 # also add SupressHTMLPreamble once we use HeaderName and some fancy CGI
-IndexIgnore .htaccess *~ .#* #*# CVS .svn favicon.ico favicon.gif robots.txt
```
domenic: robots.txt certainly exists... I guess not in any directories without a index page?

foolip: Exactly, there just one in the root. I'll update the commit message.

PR #107, opened 2017-10-23
Redirect whatwg.org/images to images.whatwg.org

foolip:
All of the common files are identical, and README is moved over.


foolip:
This is part of hosting whatwg.org, by getting rid of /images entirely there's slightly less to verify as correctly converted from .htaccess to nginx config.

domenic:
You may want to wait for @sideshowbarker's more competent review BTW; I gave you the green checkmark though to unblock you if that'd be helpful.

foolip:
> LGTM assuming that the 301 also redirects all files inside that directory.

That is what I concluded after testing with some other redirects. Seems like Apache's default is that when the 2nd argument doesn't end with a slash, it'll redirect exactly that, /images/foo, but not /imagesfoo.

What I'll do is merge this, verify that the changes are deployed and do some testing. If it didn't work as expected I'll immediately revert.

And yeah, I'm assuming that the default cache directives are good enough, since we're not doing anything special for the much bigger directory. We have monitoring in place so that we should notice a total meltdown pretty quickly.

foolip:
Deployed and confirmed to be working.

PR #110, opened 2017-10-29
Rename index to index.html on whatwg.org

foolip:
The many files on whatwg.org with no file ending require a bunch of
fiddling with content-type. Most of the URLs need to be preserved so
that renaming would require redirecting, but not so for these index
files.

If it turns out that any links are broken, redirecting foo/index to foo/
everywhere wouldn't be hard.

Note that https://whatwg.org/specs/web-workers/current-work/ (without
index.html) is redirected but
https://whatwg.org/specs/web-workers/current-work/rationale.html is not.


whatwg.org/.htaccess
```diff
@@ -107,7 +106,7 @@ redirect 301 /specs/web-apps/current-work/websrt.html https://w3c.github.io/webv
 redirect 301 /specs/web-apps/current-work/webvtt.html https://w3c.github.io/webvtt/
 redirect 301 /specs/web-apps/current-work/ https://html.spec.whatwg.org/
 redirect 301 /specs/web-forms/tests https://github.com/w3c/web-platform-tests
-redirect 301 /specs/web-workers/current-work/index https://html.spec.whatwg.org/multipage/workers.html
```
annevk: Why change this?

annevk: We don't have a redirect for `/specs/web-workers/current-work/`?

annevk: Ah, I guess the idea is that nobody linked to /index to begin with, but we redirect that way to avoid redirecting the entire directory all the time.

annevk: That is rather weird syntax-wise.

foolip: Yes, it's pretty weird. I mean to make it less weird in the nginx equivalent, by excluding just rationale.html and redirecting everything else.

annevk: I somewhat prefer it if we only redirect what we know to have been a URL that was not a 404, but I realize that's not how most redirect instructions work.

foolip: It's very easy to do just that. What's not easy is to know what URLs have existed and linked to under a path.

PR #112, opened 2017-10-30
Rename the only README (without extension) to README.txt

foolip:
This is the only extensionless file under README that isn't a perl
script, and it has some extra config associated with it. It's not an
important README nor is it easy to discover from
https://whatwg.org/demos/2008-sept/, so just rename it for simplicity.


foolip:
It causes the contents of README to show up at the bottom of https://whatwg.org/demos/2008-sept/video/. That'll go away with nginx and isn't that useful, so I dropped it here.

PR #113, opened 2017-10-30
Redirect demos that use CGI to source on GitHub

foolip:
These are linked from https://whatwg.org/demos/ as "historical
archives". This seems like a decent way to keep them around for
posterity even as CGI is disabled. (whatwg.org is being transitioned to
nginx and hosting only static content.)


foolip:
> Do you think we should also update https://whatwg.org/demos/ to link directly instead of going through redirects?

I don't have a strong opinion, avoiding the indirection doesn't seem terribly important, and keeping the original URLs means that there's just one place to update if we migrate from GitHub to something cooler in 2020. I'll change it if you like though, just say the word.

PR #118, opened 2017-12-02
wget mirror of forums.whatwg.org

annevk:
I ran the following command:

wget --mirror --page-requisites https://forums.whatwg.org/


foolip:
I ran the same command on another machine and got almost the same number of files. I'll merge this so we can iterate.

PR #121, opened 2017-12-05
Delete sid=...  forums files

foolip:
These files are are duplicates with no or minor differences, and
we need a redirect rule to handle any URLs out there anyway.


foolip:
I'm working on the nginx config now, so sure, this can stay open for a bit.

foolip:
OK, it's in https://github.com/whatwg/misc-server/pull/48. We don't know for sure if it's working until it's live anyway, so merging this to get it out of the way.

foolip:
(I did test it with forums.foolip.org as a staging server though.)

PR #126, opened 2017-12-11
Fix the styling of https://spec.whatwg.org/

foolip:
The <strong> appears to be needed.


foolip:
Yep, that worked, didn't check why :)

PR #127, opened 2017-12-11
Manually fix a few Markdown->HTML conversion errors

foolip:
Should be fixed in the conversion script, of course, later.


othermaciej:
I'm submitting the script to the whatwg.org repo: https://github.com/whatwg/whatwg.org/pull/132

Once that lands, I think the issue should move there. My best understanding of the issue is that parens after brackets should not be treated as a link if there is an intervening space. Global replace of "] (" to "] \(" as a preprocessing step should do it, I think.

PR #129, opened 2017-12-12
Editorial: "The" WHATWG

annevk:


domenic:
This is not correct English. This mismatches the singular definite article "the" with the plural "participants". Here WHATWG is used as an adjective and as such does not get a "the".

PR #132, opened 2017-12-12
Check in scripts to generate HTML versions of policies from Markdown

### LINKED ISSUE #131: Check in scripts to generate HTML versions of policies from Markdown
Check in scripts to generate HTML versions of policy documents from the Markdown versions in the SG repo.

othermaciej:
Fixes #131


domenic:
Probably all of these should be in the root since everything in the whatwg.org subdirectory gets automatically deployed to the server.

othermaciej:
OK, wasn't sure. I'll change it that way.

annevk:
What I don't understand is that https://github.com/whatwg/sg/blob/master/policy-link-mapping.txt defines a link mapping, but convert-all-policies.sh does essentially the same thing (and ends up generating the actual resources that become the links).

It seems that for a true authority split the shell script should consult the link mapping to determine what filenames to use for the output. (Of course, even in the case of a change that would likely still require updates to other parts of the site and addition of redirects and such, so it would still require coordination.)

othermaciej:
I don't expect the shell script to be used as the final build process, it's just a quick and dirty hack for my own convenience while we are doing it manually. I figured others could make use of it too. I'd be glad to make a script that determines input and output names from policy-link-mapping.txt, if such a thing would be useful for the site build process. I am not really sure how that works.

PR #135, opened 2017-12-14
Stop deploying branch snapshots

foolip:
These are not useful now that we have PR preview:
https://github.com/whatwg/meta/issues/53

This allows for some simplification of the script too.

Fixes https://github.com/whatwg/meta/issues/52.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -13,10 +13,8 @@ TITLE=$(grep < "$INPUT_FILE" "^Title: .*$" | sed -e "s/Title: //")
 
 LS_URL="https://$SHORTNAME.spec.whatwg.org/"
 COMMIT_URL_BASE="https://github.com/whatwg/$SHORTNAME/commit/"
-BRANCH_URL_BASE="https://github.com/whatwg/$SHORTNAME/tree/"
 WEB_ROOT="$SHORTNAME.spec.whatwg.org"
 COMMITS_DIR="commit-snapshots"
-BRANCHES_DIR="branch-snapshots"
 
 # Optional environment variables (won't be set for local deploys)
 TRAVIS=${TRAVIS:-false}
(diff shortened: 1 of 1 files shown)


foolip:
I've tested this with local deploy for dom and encodings to confirm the HTML output is the same module boring stuff, and for encodings all the right files still seem to be there.

resources.whatwg.org/build/deploy.sh
```diff
@@ -78,20 +77,17 @@ run_post_build_step() {
...
-         -F md-warning="Branch $BRANCH $BRANCH_URL_BASE$BRANCH replaced by $LS_URL" \
-         -F md-title="$TITLE (Branch Snapshot $BRANCH)" \
-         -F md-Text-Macro="SNAPSHOT-LINK $BACK_TO_LS_LINK" \
-         > "$BRANCH_DIR/index.html";
-    copy_extra_files "$BRANCH_DIR"
-    run_post_build_step "$BRANCH_DIR"
-    echo "Branch snapshot output to $BRANCH_DIR"
-else
-    # Commit snapshot, if master
+# Living standard
```
annevk: This used to only run in the if $BRANCH is "master". Seems better to keep it that way?

foolip: That seemed broken to me, then we will effectively do nothing useful and should instead exist early above. And doing the build and the validator steps while the PR is still open seems better.

annevk: But don't we then update the website with the contents of the PR?

foolip: No, see the change after "Deploy from Travis on master branch only"

annevk: Ah I see. I think this looks good then, but I'll let @domenic confirm.

domenic: Note that we're already, per line 33, skipping all builds for a pull request. Was your intention to change that? If not, this doesn't make too much sense.

domenic: I guess one notable change here is that local deploys will always be as-if they are master/living standard, even if they're on a branch. That's probably for the better.

resources.whatwg.org/build/deploy.sh
```diff
@@ -78,20 +77,17 @@ run_post_build_step() {
...
+curl https://api.csswg.org/bikeshed/ -f -F file=@"$INPUT_FILE" \
+     -F md-Text-Macro="SNAPSHOT-LINK $SNAPSHOT_LINK" \
+     > "$WEB_ROOT/index.html"
+copy_extra_files "$WEB_ROOT"
+run_post_build_step "$WEB_ROOT"
+echo "Living standard output to $WEB_ROOT"
+echo ""
+
+# Commit snapshot
```
domenic: If the intention is to run this for pull requests too, we should probably omit the commit snapshot, as it doesn't provide extra validation and just makes pull requests take longer. Same with the service worker.

resources.whatwg.org/build/deploy.sh
foolip: AFAICT, that check is there to skip what we consider redundant work, seen working here:
https://travis-ci.org/whatwg/fetch/builds/309802756
https://travis-ci.org/whatwg/fetch/builds/309802707

Note that we can also do that by disabling one of "Build branch updates" or "Build pull request updates" in Travis.

I'd like to make some changes here, but then the drive-by changes eclipse the original ones, so I'll simplify this and do a follow up.

foolip: Yep. Writing much more detail in the separate PR.

resources.whatwg.org/build/deploy.sh
foolip: I was rather thinking the other way around, build for PRs instead of branches. Let's discuss in the separate PR I'm sending.

foolip:
I've reduced this to something much simpler, punted the rest to https://github.com/whatwg/whatwg.org/pull/137.

annevk:
This regressed building PRs since now $WEB_ROOT no longer exists all the time.

annevk:
Should we just remove
```
echo ""
find "$WEB_ROOT" -type f -print
echo ""
```
or is there a better fix?

foolip:
Sight, this was harder than expected :) I'll fix.

foolip:
Hopeful fix in https://github.com/whatwg/whatwg.org/pull/139

PR #136, opened 2017-12-14
Thank DigitalOcean for hosting

foolip:


foolip:
Also side-channel OK'd by @zcorpan.

PR #140, opened 2017-12-18
deploy.sh: fix the condition for when to actually deploy

foolip:


foolip:
Done on the phone, so review carefully :)

annevk:
LGTM, but will let @domenic review.

PR #90, opened 2017-09-09
changed styles of home page to be responsive

o-t-w:
I noticed there was an issue for making the homepage responsive. 
https://github.com/whatwg/whatwg.org/issues/59

Issue #149, opened 2017-12-21
Update links and layouts to harmonize with revised structure


othermaciej:
Moved from https://github.com/whatwg/sg/issues/49 on behalf of @geoffcr :

I realize that the front page of the WHATWG at https://whatwg.org/ has "history" and is not likely to change, but - to a novice user - there are some confusing elements to how the links are laid out (and some of the terminology). Here are several - some significant, some less so:

**_Workstreams_**
Where do I find a list of Workstreams? Oh - Standards; we don't call them "Workstreams" anywhere, so far as I can tell. That should presumably be fixed.
(We also might want to change the "Standards" block to say "See the other **Living** Standards developed at the WHATWG".)

**_Policies_**
Where do I find Policies? Oh - if I click on FAQ, Standards, or Participate, there's also a "Policies" tab at the top. Policies deserve a link from the main page, along with Standards and FAQ. ("Working Mode" has its own front-page button, even though it's now a subset of Policies; perhaps that should be swapped out and replaced with Policies?)

**_Signing up_**
"**Participate**" takes me to a nice narrative page, but please point people to the governing Workstream Policy and IPR Policy in addition to the Code of Conduct and the Working Mode. (This strikes me as significant.) 
"**Join**" takes me straight to the repositories instead of to the participation page. That strikes me as confusing. 

**_IRC_**
Do people still use IRC? (In other words, does it merit front-page real estate?)

_**Format**_
This is just a visual nit, but it's disorienting (to me) to fly from the "Nine Colored Blocks" front page to a "Traditional Tabbed" Standards/FAQ/Policies/Participate page, or to the Living Standards in a cleaner HTML presentation. Not sure what to fix, except perhaps to ensure that there's a consistent relationship between the tabbed pages and the front page. 

**_Review Drafts_**
Will we see the latest Review Drafts posted on the Living Standard page? For example, https://html.spec.whatwg.org/multipage/ has links to the one-page version, multipage version, developer version, translations, etc. Will there also be a "Latest Review Draft" button? (I would think so; otherwise, non-technical lawyers are going to have a hard time finding it in the repository.)

domenic:
Great thoughts.

[...]

Issue #59, opened 2017-08-24
Be mobile-friendly


domenic:
We need a meta viewport tag, and possible some different header/tabbed pages styles

annevk:
I tried to shrink the design, but it actually makes it looks worse on mobile than how it renders today (which is reasonable but small, so likely hard to read for some folks). I suspect we'd need a new design and maybe a simplification of the `<h1>` as it's rather wordy for a small screen.

domenic:
Agreed. While we're redesigning, IMO it'd be nice to put the logo on each page in a corner or similar.

zouhir:
I think this is a good opportunity to slightly improve the current design, colors and overall feels into a more modern color palette and design. I'd break this into 5 tasks:

- Mobile friendly meta tags with initial scale 1.0 and no max scale so users can zoom.
- proper layout for home page which will break once we go mobile responsive.
- solution for tabbed page alternative (important to have).
- Logo on each page (@domenic  comment above)
- new website color palette (nice to have).

### Logo on each page

Do you mean just have an SVG logo instead of the large h1 on the tabbed pages?

### options for tabbed page (kind of important to have):

- off-canvas nav bar to the **left** side with full page height.
- same off-canvas nav bar but to the **right**.
- off canvas drop from the top with dynamic height (won't cover the page)

### options for colors (less important for now)

I quite like the plain white background for content rich websites like this one however we can play with the front page colors as a start.
here's some colors for inspiration:
![inspirations](https://user-images.githubusercontent.com/5052316/29714696-92467246-89e7-11e7-989d-30c38105bec6.png)

domenic:
Yeah, definitely a redesign would be welcome. In general for these sorts of things, anyone who is willing to do the work gets to pick how it looks; I don't think anyone here holds strong opinions on nav bar placement or design.

> Do you mean just have an SVG logo instead of the large h1 on the tabbed pages?

I don't think I'd get rid of the h1 entirely, but maybe it should be shrunken to just "The WHATWG" on mobile or something.

My main issue was that our beloved logo is just completely missing while you read important non-home pages like https://whatwg.org/working-mode .

annevk:
In case it was unclear, the logo is https://resources.whatwg.org/logo.svg.

zouhir:
thanks @domenic @annevk, I might be able to assist here and the yeah logo is very clear 😅  #beauty

o-t-w:
I've had a go at the frontpage. Merge if you like it

[...]

PR #153, opened 2017-12-21
Redesign the homepage

domenic:
Closes #90 by subsuming the work there. Helps with #59 as at least the homepage is now mobile-friendly. Helps with some of the items noted in #149.

Substantive changes:

* "Join" renamed to "GitHub" (see #149)
* "News" renamed to "Blog"
* "Working Mode" replaced with "Policies" (see #149)
* "Twitter" added

---

Thank you @o-t-w for your original design in #90! This is all-new code, but based on your visual ideas.


domenic:
I forgot the most important part: a preview! Most of the links will not work (because I haven't set up a full nginx server to handle our extension-less files), but you can see the result at https://whatwgorg-wjjgptrgid.now.sh

domenic:
I guess this isn't quite mergeable as-is because /images currently is an alias for images.whatwg.org. I'm not sure what our plan is there; thoughts appreciated.

annevk:
I don't think we should make images/ an alias for images.whatwg.org if we need whatwg.org-specific images. We should just dedupe the resources and that's that.

annevk:
Couple design points:

* Heading could use more space at the top.
* Grey looks rather plain.
* Should experiment a bit more with font weights and different fonts if you haven't already. Also looks rather plain at the moment.
* I wonder if we should give the footer have a normal font size to make it a bit more readable.

o-t-w:
Sorry for going AWOL. I was moving house and changing jobs at the same time and feeling very stressed!
- I agree the top definitely needs slightly more padding. 
- I like the grey - its undistracting. 
- I always value the speed/performance of a website over visual flourishes, which is why I didn't use a font that the users browser will need to download. The majority of free fonts in the world are utter garbage, and licensing a decent one can be prohibitively expensive. But I agree the current font is slightly bland. Using system fonts can prevent the extra HTTP request while still looking more stylish: `html {font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif }`
- Font-sizing is important to establishing a visual hierarchy.  The small size of the type in the footer denotes that it is of less importance - its less likely to be information that visitors would be looking for. It might be good to change the font-sizing of the footer to use em units instead of pixels so that user browser-settings for font-size are respected. That way it is accessible for people with bad eyesite, while still respecting the design.

domenic:
> I don't think we should make images/ an alias for images.whatwg.org if we need whatwg.org-specific images. We should just dedupe the resources and that's that.

The issue is that it's already at 301 redirect, so I'm not sure how to proceed.

> Heading could use more space at the top.
>
> Should experiment a bit more with font weights and different fonts if you haven't already. Also looks rather plain at the moment.

In general with this and other design points, I'd rather land then iterate, or at least get very specific suggestions; I don't feel equipped to respond to feedback like this myself.

> It might be good to change the font-sizing of the footer to use em units instead of pixels so that user browser-settings for font-size are respected.

My understanding is that these days pixels vs. ems make no difference in this regard.

foolip:
> The issue is that it's already at 301 redirect, so I'm not sure how to proceed.

How about make it a 302 and wait for at least a week, or as long as it takes for this PR to be done?

foolip:
Actually, per https://stackoverflow.com/a/21396547 that might not work. Maybe we should just stop using permanent redirects entirely, except for HTTP->HTTPS?

domenic:
It's also unclear to me how often people might hit those URLs. E.g. I imagine a lot of the old specs include /images URLs.

Maybe we should create a new directory, e.g. `/img`?

foolip:
Yes, /images is linked a lot from specs/, but also demos/, news/ https://whatwg.org/style/tabbed-pages.

If we did ram through the redirect caching we'd still end up with a bunch of old cruft in the directory, so I don't have a strong opinion, whatever works.

foolip:
> Yeah I'm not sure why we made it a redirect.

That was me in commit 95959905618cdbf9ac44f33e9fca572872d4c606, now it's clear the other way around would have been better.

I've sent https://github.com/whatwg/misc-server/pull/66 to make this less painful if we change it.

domenic:
Updated based on feedback; new preview at https://whatwgorg-ymcoicnlew.now.sh

domenic:
Oops, typoed the extra spacing, see https://whatwgorg-llsduqszfy.now.sh. (It now matches @o-t-w's original.)

foolip:
I think this looks pretty good!

The amount of padding above/below the logo/title looks a bit uneven to me, can we try a bit more on top?

In the dark grey boxes starting with "Participate", a bit more horizontal padding would be nice, maybe not as much as the vertical, but a tad. At least the width of the lower case "y" or "s" the the strings end with.

The green gradients aren't bad, certainly better than solid WHATWG green, but a solid `#00981B` would also look OK. What do others think?

domenic:
I'd prefer concrete CSS suggestions to "how about maybe changing this a bit?" :). I don't feel qualified to translate such things.

annevk:
Here is some CSS that makes it look a little better and gives the links hover/focus effects. Not entirely sure we also need visited styling here. Probably okay without...
```css
* {
 box-sizing: border-box;
}

:link {
 color: #00F;
}
:visited {
 color: #0E388C;
}

html {
 font-family: -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, Helvetica, Arial, sans-serif;
 background-color: rgb(235, 235, 235);
 --whatwg-green: #3A7908;
}

body {
 max-width: 900px;
 margin: 0 auto;
 padding: 0;
 display: flex;
 flex-direction: column;
 min-height: 100vh;
}

body > *:not(footer):not(script) {
 margin-bottom: 40px;
}

header {
 margin:2em 0;
 display: flex;
 align-items: center;
}

header > hgroup {
 margin-left: 20px;
}

header > hgroup > h1 {
 margin: 0;
 font-weight: normal;
}

@media (max-width: 600px) {
 header > hgroup > h1 {
  font-size: 19px;
 }
}

header > hgroup > h2 {
 font-size: 17px;
 margin: 0;
 color: rgb(40, 40, 40);
 font-weight: normal;
 margin: 5px 0;
}

#the-standards > ul {
 list-style-type: none;
 margin: 0;
 padding: 0;
}

#the-standards li {
 border: solid var(--whatwg-green) 2px;
 border-radius: 7px;
 letter-spacing: 0.25px;
 font-weight: bold;
 background-color: white;
}

#the-standards li:not(:last-of-type) {
 margin-bottom: 15px;
}

#the-standards a {
 color: var(--whatwg-green);
 text-decoration: none;
 padding: 4px 30px;
 height: 64px;
 display: flex;
 align-items: center;
}

#the-standards a:hover, #the-standards a:focus {
  transition:.2s; 
  padding-left:50px;
}

#the-standards a::before {
 display: inline-block;
 width: 27px;
 height: 15.4px;
 background: url(/img/arrow.svg);
 content: "";
 margin-right: 21px;
 vertical-align: middle;
 background-repeat: no-repeat;
}

@media (min-width: 600px) {
 #self-explanatory-links {
  display: grid;
  grid-template-columns: repeat(12, 1fr);
  grid-column-gap: 20px;
  grid-row-gap: 20px;
 }
}

#self-explanatory-links > a {
 padding: 10px;
 text-align: center;
 border: 8px solid rgba(58, 121, 8, 0.2);
 background: linear-gradient(var(--whatwg-green), #00981B);
 background-clip: padding-box;
 grid-column: span 3;
 text-decoration: none;
 color: white;
 font-size: 17px;
 letter-spacing: 0.3px;
}

#self-explanatory-links > a:hover, #self-explanatory-links > a:focus {
  transition:.2s;
  transform:scale(1.1);
}

@media (max-width: 600px) {
 #self-explanatory-links > a {
  display: flex;
  align-items: center;
  justify-content: center;
  flex-direction: column;
  margin-top: 15px;
 }
}

#links-with-explanations {
 /* Grow into the extra space before the footer to make the footer stick to the bottom. */
 flex: 1;
}

#links-with-explanations a {
 padding:0 10px;
 display: flex;
 flex-wrap: wrap;
 margin-bottom: 15px;
 background-color: #333;
 align-items: center;
 justify-content: space-between;
 border-radius: 4px;
 color: white;
 min-height: 54px;      
 text-decoration: none;
}

#links-with-explanations a:hover, #links-with-explanations a:focus {
  transition:.2s;
  background-color: #000;
}

#links-with-explanations a::before {
 display: inline-block;
 width: 30px;
 height: 30px;
 content: "";
 margin: 12px 20px 12px 8px;
 vertical-align: middle;
 background-repeat: no-repeat;
}

#participate::before {
 background: url(/img/participate.svg);
}

#irc::before {
 background: url(/img/irc.svg);
}

#twitter::before {
 background: url(/img/twitter.svg);
}

#checker::before {
 background: url(/img/checker.svg);
}

#links-with-explanations a > strong {
 margin-right: auto;
}

#links-with-explanations a > p {
 margin-left: auto;
 padding-right: 8px;
}

@media (max-width: 600px) {
  #links-with-explanations a > p {
    width: 100%;
    padding-left: 12px;
  }
}

footer {
 border-top: 1px solid black;
 text-align: center;
 font-size: 12px;
}

footer small {
  /* Semantically it's small, but the footer is already small, so don't change the font size. */
  font-size: inherit;
}

footer > * {
 font-style: normal;
 margin: 12px 0;
}
```

domenic:
Alrighty, applied those plus some further small tweaks for mobile. Thanks especially for the hover states; those were conspicuously missing.

Thoughts? I'll take a break from trying to work on this and #154 in tandem; let's land this first then I can work on that on top of it.

foolip:
@domenic, is there a preview URL?

domenic:
I made one but forgot to paste it: https://whatwgorg-ypuhyslfsp.now.sh

domenic:
I think of it more as a variable name than a color name, but shrug.

I'll give people some more time to check it out before merging. Woohoo!

PR #156, opened 2018-01-04
Keep the margin on <p>s inside notes

domenic:
Before:  after:


annevk:
It looks better there, but I wonder if there's a case in the HTML Standard that justifies this being added to begin with.

domenic:
I did a random spot check of 6 of the 55 occurences of the first-child case in HTML and couldn't find any difference. I think we should merge and remember this if something looks weird.

PR #158, opened 2018-01-04
Fix deploy script for recent Bikeshed boilerplate updates

domenic:
- We don't use Title: in the source documents anymore, so scrape the H1: instead
- Work around https://github.com/tabatkins/bikeshed/issues/1162

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -9,7 +9,7 @@ set -o pipefail
 
 SHORTNAME=$(git config --local remote.origin.url | sed -n "s#.*/\([^.]*\)\.git#\1#p")
 INPUT_FILE=$(find . -maxdepth 1 -name "*.bs" -print -quit)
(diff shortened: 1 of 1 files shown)


domenic:
tabatkins/bikeshed#1162 has been fixed but it seems like elsewhere in the header we prefer to omit quotes around attribute values so I'll leave that change in anyway.

resources.whatwg.org/build/deploy.sh
```diff
@@ -76,7 +76,7 @@ echo ""
 # Living standard
 curl https://api.csswg.org/bikeshed/ -f -F file=@"$INPUT_FILE" \
      -F md-Text-Macro="SNAPSHOT-LINK $SNAPSHOT_LINK" \
-     > "$WEB_ROOT/index.html"
+     > "$WEB_ROOT/index.html";
```
annevk: Nothing else here uses ";". Was this change intentional?

Issue #117, opened 2017-11-30
Build script improvements


annevk:
I'd like the build script to also report any Bikeshed WARNING and LINK ERROR messages, possibly as fatal. Otherwise we risk introducing subtle issues and letting it all get worse over time. https://github.com/whatwg/fetch/issues/639 is a good example of that happening. Maybe some standards will always have WARNING or LINK ERROR messages, but it seems to me we should not tolerate them. (I actually thought this was fixed for Encoding, but it seems to still happen.)

And in general it should do better with error handling. For FATAL ERROR messages you currently don't see anything and have to run Bikeshed again yourself.

Instead if upon failure we first run `curl https://api.csswg.org/bikeshed/ -f -F file=@fetch.bs -F output=err -F md-Text-Macro="SNAPSHOT-LINK LOCAL COPY"` or some such or store the errors from the initial run directly somehow and then fail the build we'd at least have something to work with.

domenic:
This is basically a dupe of https://github.com/whatwg/meta/issues/24; not sure which to close.

annevk:
We can have them both until it's fixed. Both are a natural place to search for this issue.

annevk:
Another thing: @foolip and have noticed connection issues now and then. I just got hit by "Failed to connect to api.csswg.org port 443: Connection timed out" and we've got the same error for resources.whatwg.org. We could use https://docs.travis-ci.com/user/common-build-problems/#Timeouts-installing-dependencies to make Travis try again once or twice to not have to hit retry ourselves.

annevk:
The current situation is really bad. I've basically no idea what's going with Infra at the moment:
```
$ curl --remote-name --fail https://resources.whatwg.org/build/deploy.sh && bash ./deploy.sh
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100  4546  100  4546    0     0  21753      0 --:--:-- --:--:-- --:--:-- 21855
Commit = af5f4d000d4568fd060b7ff9a450c900a711e773
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 49857    0     0  100 49857      0  42841  0:00:01  0:00:01 --:--:-- 42869
curl: (22) The requested URL returned error: 500 Internal Server Error
[...]

PR #160, opened 2018-01-06
Check for Bikeshed errors and warnings during deploy

annevk:
Also add even more context and improve the context added in a prior commit somewhat.

Helps with #117.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -96,7 +96,7 @@ importScripts(\"https://resources.whatwg.org/standard-service-worker.js\");
 echo "Service worker script output to $WEB_ROOT"
 echo ""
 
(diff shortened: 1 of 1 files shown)


annevk:
With this merged it seems we could rerun the latest for each spec on master, fix any Bikeshed errors and then somehow change this script to exit early if any Bikeshed errors are found, right?

cc @sideshowbarker

resources.whatwg.org/build/deploy.sh
```diff
@@ -93,13 +96,16 @@ importScripts(\"https://resources.whatwg.org/standard-service-worker.js\");
 echo "Service worker script output to $WEB_ROOT"
 echo ""
 
+echo "Overview of generated files..."
```
domenic: Nit: no "..." here as it isn't a process that takes time

Issue #80, opened 2017-09-01
Updating description of whatwg@whatwg.org


annevk:
In particular we should probably mention some of the more valuable ways to participate and mention that the list is primarily meant as a fallback and maybe for announcements. And I suppose whatever purpose people want to use it for that isn't too upsetting to the 1000 or so subscribers.

We should also not write "WHAT Working Group".

Anyone more thoughts?

PR #164, opened 2018-01-07
Advise people to use GitHub over the mailing list

### LINKED ISSUE #51: Move mailing list related wiki FAQ entries to /mailing-list
> Not necessarily in this PR, but I realized that the remaining content of https://wiki.whatwg.org/wiki/FAQ could probably go here.

annevk:
Also inline the remaining mailing list FAQ entries from the wiki.

Fixes #51.

### DIFF
--- a/whatwg.org/mailing-list
+++ b/whatwg.org/mailing-list
@@ -31,7 +31,8 @@
   HTML5?", then please use <a href="https://stackoverflow.com/">Stack Overflow</a>.</p>
 
   <p>It is preferable to <a href="https://github.com/whatwg">interact with the WHATWG community
-  through GitHub</a>, but occasional announcements and discussion will also happen on
+  through GitHub</a>, with the <a href="https://github.com/whatwg/meta">whatwg/meta repository</a>
+  as the general discussion place, but occasional announcements and discussion will also happen on
   whatwg@whatwg.org.
 
   <h3>Archives</h3>
--- a/whatwg.org/mailing-list
+++ b/whatwg.org/mailing-list
@@ -24,62 +24,102 @@
    <li><a href="https://participate.whatwg.org/">Participate</a></li>
   </ul>
 
-  <h2>Mailing List</h2>
+  <h2 id=spec>Mailing List</h2>
 
   <p>Do you have questions on how to use HTML5? If you want to ask "how do I upgrade my HTML4 page
   to use Web Forms?", "what is a web application?", "why does this not work?", "how do I use
   HTML5?", then please use <a href="https://stackoverflow.com/">Stack Overflow</a>.</p>
 
-  <div id="spec">
+  <p>It is preferable to <a href="https://github.com/whatwg">interact with the WHATWG community
+  through GitHub</a>, but occasional announcements and discussion will also happen on
+  whatwg@whatwg.org.
 
-   <h2 id="specs">The specifications</h2>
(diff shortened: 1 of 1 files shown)


annevk:
This also fixes #80.

annevk:
Note that post-landing this https://wiki.whatwg.org/wiki/FAQ needs to be updated.

whatwg.org/mailing-list
```diff
@@ -24,62 +24,102 @@
...
 
-  <h2>Mailing List</h2>
+  <h2 id=spec>Mailing List</h2>
 
   <p>Do you have questions on how to use HTML5? If you want to ask "how do I upgrade my HTML4 page
   to use Web Forms?", "what is a web application?", "why does this not work?", "how do I use
   HTML5?", then please use <a href="https://stackoverflow.com/">Stack Overflow</a>.</p>
 
-  <div id="spec">
+  <p>It is preferable to <a href="https://github.com/whatwg">interact with the WHATWG community
```
domenic: Should we mention whatwg/meta as a general discussion place?

whatwg.org/mailing-list
```diff
@@ -24,62 +24,102 @@
...
+  moderation queue, it means it has been caught by our spam filter. Make sure you subscribed and
+  posted using the right email address and make sure you didn't send any attachments. If you still
+  can't get your email sent to the list, please open a
+  <a href="https://github.com/whatwg/meta/issues/new">new meta issue</a>.</p>
 
-   <p>To unsubscribe from the mailing list, send an email
-   to <a href="mailto:whatwg-unsubscribe@whatwg.org">whatwg-unsubscribe@whatwg.org</a>
-   from the address you want to unsubscribe, and then reply to the
-   confirmation email you receive.
+  <h3>Can I use +1?</h3>
```
domenic: I was wondering if we should copy this to the FAQ "how to interact on GitHub". Although, I think most people on GitHub have gotten the message to use the GitHub reactions feature instead of +1 messages, so maybe it's not necessary.

whatwg.org/mailing-list
```diff
@@ -24,62 +24,102 @@
...
+  <p><strong>Please reply inline or make the reply self-contained, and trim extraneous quotes from
+  previous emails in your replies.</strong></p>
+
+  <p>Basically, please remove anything after the last line you have written, so that people don't
+  have to scroll down to find out what else you wrote, and make sure that your email makes sense on
+  its own, as it will probably be read out of context years later.</p>
+
+  <p>That is, you should reply like this:</p>
+
+  <pre>Ian wrote:
```
domenic: Blockquote these (in addition to pre)?

whatwg.org/mailing-list
michaelchampion: I think it's worth making the advice explicit -- pose questions as suggested positions, then expect people to give an appropriate reaction.  For example, this question might be posed: "I think most people on GitHub have gotten the message to use the GitHub reactions feature instead of +1 messages, so it's not necessary.  Agree?"  Then I would have just given it a thumbs up rather than typing ;-)

annevk:
1. I added whatwg/meta. 
2. I don't think we need to mention it for GitHub since it generally doesn't happen (although I see it occasionally still). If it happens more we can add it.
3. It's not a quote, just an example. Do we have `<div class=example>`?

PR #165, opened 2018-01-07
Update n.whatwg.org namespace documents

### LINKED ISSUE #99: Add a page at n.whatwg.org/work
It should link to https://html.spec.whatwg.org/#licensing-works (or the multipage variant) and should clarify that despite web browsers redirecting to https://, for the purposes of the namespace you must use http://.

annevk:
Fixes #99.

### DIFF
--- a/n.whatwg.org/work
+++ b/n.whatwg.org/work
@@ -1,0 +1,6 @@
+The string http://n.whatwg.org/work (note the lack of HTTPS compared to
+the URL of this document) represents the namespace for a work. See
+
+  https://html.spec.whatwg.org/multipage/microdata.html#licensing-works
+
+for details.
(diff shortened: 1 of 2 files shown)


annevk:
This requires a corresponding fix on the server to remove the redirect.

cc @foolip @sideshowbarker

n.whatwg.org/work
```diff
@@ -0,0 +1,6 @@
+The string http://n.whatwg.org/work (note the lack of HTTPS compared to
+the URL of this document) represents the namespace for a work. See
+
+  https://html.spec.whatwg.org/multipage/microdata.html#licensing-works
```
foolip: Is the indentation intentional, given that it was removed in the other file?

foolip: Ah, it's now two spaces in both files.

domenic:
FWIW when I field #99 I think I must have been unaware of the redirect... The redirect is not too bad IMO. But if by filing #99 I've convinced others that a dedicated pointer page is better, that works too :)

foolip:
I'm fine either way, perhaps with a weak preference for what Anne has done here.

PR #166, opened 2018-01-08
Make standards deploy a robots.txt resource too

annevk:
Fixes https://github.com/whatwg/meta/issues/32.


annevk:
Works perfectly locally (and was based on the service worker work).

foolip:
> was based on the service worker

Oh, didn't notice the similarity :)

Issue #133, opened 2017-12-13
Automate generating whatwg.org policy pages from whatwg/sg


domenic:
The script in #132 is used to generate several pages on whatwg.org from Markdown documents in whatwg/sg. We should figure out a way to automate this.

Several possibilities:

1. Build process lives here in whatwg/whatwg.org's Travis job; pushes to whatwg/sg trigger a whatwg/sg Travis job which uses https://docs.travis-ci.com/user/triggering-builds to trigger the whatwg/whatwg.org Travis job.
2. The Travis job for whatwg/sg could do the build, then commit and push the generated output into whatwg/whatwg.org directly, thus indirectly triggering the whatwg/whatwg.org Travis job (which just does a deploy).
3. The Travis job for whatwg/sg pushes a small file containing a commit hash for whatwg/sg into whatwg/whatwg.org, and the whatwg/whatwg.org build process uses whatever hash is committed to build. (A variant on (1).)

I like (1) the most.

annevk:
Reading triggering-builds it sounds like whatwg/sg needs to get a private key and associate that with Travis. That key needs to encrypt the whatwg/whatwg.org token and then the whatwg/sg build script decrypts it and uses it to trigger the whatwg/whatwg.org build.

That way we can use Travis for everything and don't need a script on a server that's triggered through a webhook. Does that sound accurate? I can work on that, but I think @domenic can probably pull it off a little quicker.

annevk:
https://hiddentao.com/archives/2016/08/29/triggering-travis-ci-build-from-another-projects-build/ documents this though it uses Node.js when a simple curl would hopefully do.

annevk:
Given the low volume of changes doing this manually has been fine.

PR #167, opened 2018-01-08
Automatically import content from whatwg/sg

annevk:
Fixes part of #133.


annevk:
1. When I tried to run this locally it failed because I don't have `markdown` installed. Do we know what Travis CI has installed?
2. The current structure of these scripts is rather weird as we have two places that dictate the URLs of these files. This script and https://github.com/whatwg/sg/blob/master/policy-link-mapping.txt. Both end up being used.
3. Whenever we add more documents we'll need updates in both places. I don't think that's really problematic and probably helps preventing accidental errors.

annevk:
cc @othermaciej

annevk:
(The other thing that bugs me is that there's lots of repetition in the setup. It'd be nicer to iterate over a map of sorts and then invoke.)

deploy.sh
```diff
@@ -4,6 +4,16 @@ set -e
 if [[ "$TRAVIS_BRANCH" != "master" || "$TRAVIS_PULL_REQUEST" != "false" ]]; then
     echo "Skipping deploy for a pull request"
 else
+    echo "Importing content from whatwg/sg and making it suitable for whatwg.org"
```
domenic: Seems like this should happen either way (outside the if/else) so that we can test the conversion when people submit changes to the script.

othermaciej:
@annevk I could make a single Python script that reads from policy-link-mappings.txt and figures out to convert based on that, if it would help. Maybe after this lands, since it can easily be done as a follow-on change.

deploy.sh
othermaciej: The correct invocation is `pip install markdown py-gfm`

annevk:
@othermaciej it seems that would be cleaner, yes. Happy as a follow-up.

deploy.sh
annevk: @domenic how would that actually test the conversion though? Or do you mean locally? Locally seems reasonable, but I'd like to avoid a git clone for each PR.

annevk:
I think this should work. It works locally with that pip install command (though I think my local setup is somewhat ruined as I had to use sudo).

deploy.sh
```diff
@@ -30,4 +27,5 @@ else
     ssh-add deploy_key
     echo "$SERVER $SERVER_PUBLIC_KEY" > known_hosts
     rsync --archive --chmod="D755,F644" --verbose --compress --delete --rsh="ssh -o UserKnownHostsFile=known_hosts" ./whatwg.org ./*.whatwg.org "deploy@$SERVER:/var/www/"
+    echo ""
```
domenic: I'm surprised this is necessary/helpful?

annevk: I'm not actually sure, I'll remove it for now.

annevk:
Seems this worked out fine. \o/

PR #171, opened 2018-01-09
Further improvements to standard build logs

annevk:
In particular, make "headers" stand out more.


annevk:
Without the headers the curl stuff was rather hard to follow to me. And without this the upcoming linting stuff is somewhat hard to follow, though it's not too bad. But I'm also playing a bit with this to learn. I actually hope the end result is that I don't have to look much at these logs anymore as all the important information ends up in a GitHub comment.

PR #172, opened 2018-01-09
Lint build wip

annevk:


annevk:
This works (tested with fetch), but doesn't have a good commit message yet. Ready for review I think.

annevk:
Commit:
`Add lint checking similar to whatwg/html-build to other standards

While this results in some duplication with whatwg/html-build, it's rather hard to avoid doing that and the benefits outweigh the drawbacks.

Fixes ##66.`

resources.whatwg.org/build/deploy.sh
```diff
@@ -64,6 +63,22 @@ header() {
   echo -e "\x1b[1m$1\x1b[0m"
 }
 
+header "Linting the source:"
+MATCHES=$(
+  grep -niE '( (code|span|var)(>| data-x=)|[^<;]/(code|span|var)>)' "$INPUT_FILE" | perl -lpe 'print "\nPossible copypasta:" if $. == 1'
```
zcorpan: data-x is specific to HTML spec, but I guess that's ok.  This check is trying to catch missing the "<" in start or end tag, because the messages from the validator don't point to the cause.

PR #175, opened 2018-01-10
Retry most curl requests as they sometimes timeout

### LINKED ISSUE #117: Build script improvements

annevk:
We are not a 100% sure what exactly causes them to timeout as locally they always seem available around the same time, but hopefully this reduces the fail rate of PRs a bit.

Note that this does not address curl'ing of the deploy.sh script itself and also does not touch curl'ing the HTML checker as that code is likely to change anyway.

Fixes #117 (the remainder).

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -64,28 +64,34 @@ header() {
   echo -e "\x1b[1m$1\x1b[0m"
 }
 
+curlretry() {
+  curl --fail --retry 2 "$@"
+}
+
+bikeshed() {
+  curlretry https://api.csswg.org/bikeshed/ -F file=@"$INPUT_FILE" "$@"
+}
+
 header "Checking for errors and warnings using Bikeshed..."
-curl https://api.csswg.org/bikeshed/ -f -F file=@"$INPUT_FILE" -F output=err \
-     -F md-Text-Macro="SNAPSHOT-LINK ERROR WARNING CHECK";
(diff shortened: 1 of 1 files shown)


resources.whatwg.org/build/deploy.sh
```diff
@@ -64,36 +64,42 @@ header() {
   echo -e "\x1b[1m$1\x1b[0m"
 }
 
+curlretry() {
+  curl --fail --retry 2 "$@"
+}
+
+bikeshed() {
```
domenic: Maybe for clarity "curlbikeshed" or "remotebikeshed"? Otherwise it's too easy to be reading things and thinking we've somehow locally installed bikeshed.

PR #177, opened 2018-01-10
Remove obsolete code from convert-policy.py

annevk:
This doesn't do anything due to a typo, but is also no longer needed as the titles are now consistent upstream (whatwg/sg).


annevk:
@othermaciej ping.

PR #183, opened 2018-01-30
Update Fullscreen "authors" in biblio.json

foolip:
Matches https://github.com/whatwg/sg/pull/60.


foolip:
Is this what you had in mind? Is there any logic to when we list multiple editors and not?

foolip:
OK, I might still have to send you reviews though :)

PR #189, opened 2018-03-08
Make padding consistent for Note and Example sections

### LINKED ISSUE #187: [standard.css] Make padding consistent
There are several elements in the WHATWG spec that could benefit from added padding (specifically padding-top, padding-bottom, and padding-right). Right now text runs along the edge of these elements, which isn't exactly visually appealing.

Here's a list of the elements I found that benefit from padding:

- Note (`.note`)
- Example (`.example`)

Other classes like `.domintro` and `.idl` already make use of padding.

I'd like to submit a pull request fixing this if possible.

GloverDonovan:
This is just a quick fix to the padding issue for the Note and Example sections.

I chose not to use CSS variables here because:

- I'm not sure how many people still view WHATWG specs with older browsers. Don't want to break anyone's setup with a trivial change.
- Just having a few variables would be inconsistent IMO. A rewrite is better here.

The stylesheet could be a lot more organized and tidy (more comments, line breaks, etc.), but that's another issue.

Fixes #187.

<details><summary>Screenshots</summary>
<br>
Before:

<br>
After:

<br>
Before:

<br>
After:

</details>

### DIFF
--- a/resources.whatwg.org/standard.css
+++ b/resources.whatwg.org/standard.css
@@ -155,7 +155,7 @@ dl.domintro dd { margin: 0.5em 0 1em 2em; padding: 0; }
 dl.domintro dd p { margin: 0.5em 0; }
 dl.domintro::before { content: 'For web developers (non-normative)'; background: green; color: white; padding: 0.15em 0.25em; font-style: normal; position: absolute; top: -0.8em; left: -0.8em; }
 
-.note { position: relative; color: green; background: #DDFFDD; font-style: italic; margin-left: 2em; padding-left: 2em; }
+.note { position: relative; color: green; background: #DDFFDD; font-style: italic; margin-left: 2em; padding: 0.5em 1em 0.5em 2em; }
 .note em, .note i, .note var { font-style: normal; }
 span.note { padding: 0 2em; }
 dd > .note:first-child { margin-bottom: 0; }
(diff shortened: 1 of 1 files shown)


resources.whatwg.org/standard.css
```diff
@@ -167,8 +167,10 @@ dd > .note:first-child { margin-bottom: 0; }
 .warning p:last-child { margin-bottom: 0; }
 .warning::before { font-style: normal; content: '\26A0 Warning!'; background: red; color: yellow; padding: 0.15em 0.25em; font-style: normal; position: absolute; top: -0.2em; left: -4.25em; transform: rotate(-5deg); }
 
-.example { display: block; color: #222222; background: #EEEEEE; margin-left: 2em; padding-left: 3em; position: relative; }
+.example { display: block; color: #222222; background: #EEEEEE; margin-left: 2em; padding: 0.5em 1em 0.5em 3em; position: relative; }
 .example::before { font-style: normal; content: 'Example'; background: #222222; color: #EEEEEE; padding: 0.15em 0.25em; font: 1em Helvetica Neue, sans-serif, Droid Sans Fallback; position: absolute; top: 0.2em; left: -2.25em; transform: rotate(-5deg); }
+.example p:first-child { margin-top: 0; }
+.example p:last-child { margin-bottom: 0; }
```
domenic: This is the kind of stuff I find confusing with the current system. We're turning off the margin of the first/last element, but only if the element is a p, and then we're adding padding instead? But only on warnings and examples; notes don't have this first/last child thing? Because I removed it in https://github.com/whatwg/whatwg.org/pull/156, I think because of the case of the note after https://streams.spec.whatwg.org/#rbs-controller-internal-slots.

I propose something a bit more drastic: use 1.25em vertical margin on .note/.warning/.example (to match the p margin); use the 0.5em vertical padding you have here; and don't include any first-child/last-child margin removal (so get rid of it on .warning), but instead rely on margin-collapsing.

Would you be up for trying that, and seeing what the impact is? Important cases to keep in mind are bare div.note with no p inside, div.note with a single p inside, and div.note with multiple ps inside. I think the above makes them all work.

GloverDonovan: I believe that one of the problems here is using the same class (`.note`, etc.) for two completely different elements (`<div>` and `<p>`). Some specs use `<p class="note">` and other specs use `<div>` with no `<p>` tag inside them.

I looked at the section you mentioned and noticed that the table does not have a margin (more specifically `margin-top` and `margin-bottom`), but neither do notes. In both cases, one relies on the margin of the `<p>` tag, but I don't think this is ideal since the margin of the `<p>` tag goes outside of the box model of its parent container.



Read the rest of your comment and agree that notes and other sections should have their own margin (this was already the case for sections that used `<p class="...">` instead of `<div class="...">`). The `.warn` class already has a `margin-top` and `margin-bottom`, but with `2em` instead of `1.25em`.

PR #191, opened 2018-03-20
Rename logo-mime.* to logo-mimesniff.*

### LINKED ISSUE #190: Rename logo-mime.{svg,png} to logo-mimesniff.{svg,png}
(and maybe add a redirect?)

It looks like the logo for mimesniff has been broken for a month or so ever since we used the new spec template that assumes the logo name is the same as the spec shortname. Oops.

Reported by @othermaciej in IRC.

annevk:
This makes it consistent with the Bikeshed template used for the MIME Sniffing Standard and avoids us having to special case things there.

Fixes #190.


annevk:
Redirects would have to be added by @foolip or @sideshowbarker.

Issue #176, opened 2018-01-10
Making lint checks and Bikeshed warnings fatal


annevk:
Once #172 has landed we have enabled lint checking and each standard using the shared build script will output the results of those checks. They already report Bikeshed warnings.

We need to investigate those warnings for each standard and see where they need to be addressed and then address them.

Then once all standards no longer report warnings we need to make them fatal so they cannot be introduced in the future.

An alternative here is that we provide a `.travis.yml` flag so you can opt-in to strict warnings to make this transition easier, but I'd rather just go all the way.

annevk:
* MIME Sniffing is still in a bad state here
* XMLHttpRequest has a false-positive on spelling that I'm not sure how to remove

domenic:
What's the XHR false positive?

We could use a comment strategy like we did for en-GB.

annevk:
It does not like my example `unicorns-are-teh-awesome.txt`.

domenic:
The new "Informative classes" thing is creating new warnings in various standards now that domintros are properly classed as non-normative.

domenic:
https://github.com/tabatkins/bikeshed/issues/1329 is now implemented. We should audit all existing warnings and see if we can just flip the flag and fix them as we go, or need something more complicated.

annevk:
@plinss could you add support for --die-on to https://api.csswg.org/bikeshed/ pretty please?

plinss:
Added

annevk:
If you use `output=err`, would this parameter affect the output? I've intentionally not passed it to that particular Bikeshed call in my PR, but it would simplify things a bit if it didn't matter and I could pass it there.

plinss:
This parameter is orthogonal to output. `output=err` gives you the stdout and stderr output of bikeshed (with the generated output going to a temp file). The `--die-on` only changes what bikeshed considers to be a fatal error, which IIRC only affects the generated output (though it may stop on the first fatal error?).

tabatkins:
It tries to exit with a non-zero exit code, yeah.

PR #192, opened 2018-03-20
Add RGB and RST_STREAM to "an" list

annevk:
Fixes part of #176.


annevk:
I think we can start enabling hard failure as well, though it would be nice if we returned failure after running both "Linting the source:" and "Checking for errors and warnings using Bikeshed..." so you get all the possible errors at once.

Any thoughts on how to go about that? And any concerns?

foolip:
No concerns assuming that we're error-free at the outset. Do you have a way of running the check for all specs to see?

annevk:
I don't. I'm also not sure how to make it work like I described above.

Issue #162, opened 2018-01-06
Consider moving more wiki pages to whatwg.org


annevk:
Here are the ones that come to mind:

- [x] https://wiki.whatwg.org/wiki/IRC
- [x] https://wiki.whatwg.org/wiki/Style

The one problem I have is that I'm not sure what a good entry point for these pages is within the overall site hierarchy.

foolip:
Redirecting from the wiki is also not entirely straightforward. It could be, of course, but would be done by Apache.

annevk:
Thus far we used a template of sorts: https://wiki.whatwg.org/wiki/FAQ. That seems fine to me.

cc @GPHemsley

domenic:
See also #51.

GPHemsley:
We have two templates:

* https://wiki.whatwg.org/wiki/Template:Obsolete (created in 2012 by me)
* https://wiki.whatwg.org/wiki/Template:Moved (created in 2017 by @zcorpan)

The obsolete template also adds pages to various categories, rooted here: https://wiki.whatwg.org/wiki/Category:Obsolete_documents

There is indeed no convenient way to redirect to an external URL from within MediaWiki, though I think this is actually by design.

GloverDonovan:
If this is still an issue, I can work on moving the style guide (maybe it should read more naturally as well?). Who is the target audience? It may be useful to take inspiration from MDN's [Writing style guide](https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Writing_style_guide).

annevk:
@GloverDonovan the audience is folks doing any kind of writing for the WHATWG, including standards, documentation, comments in software, the website, and blog posts. Of those, the primary focus would be the standards and the website.

annevk:
I'd be happy to see it made more accessible, for what it's worth.

annevk:
Closing as other pages that need to be moved (if any) can be done through new issues.

annevk:
Thanks again @GloverDonovan for your efforts here!

GloverDonovan:
A bit late, but glad I could help!

PR #193, opened 2018-03-26
Move IRC wiki page to whatwg.org

GloverDonovan:
Hope I'm doing this right.

This is a direct copy of the IRC wiki page, but with one exception: I changed `irc://` to `ircs://` and added a colon before the `pre` block.

Ref #162


annevk:
This will also require changes to other pages that link to the IRC page, such as the home page.

GloverDonovan:
Decided to fix the IRC links directly in this pull request since there weren't as many other old links as I thought. Also updated the first commit to escape the `<` and `>` in the `pre` block.

annevk:
Updated wiki.whatwg.org and blog.whatwg.org to reflect this change as well.

domenic:
Looks really nice!

PR #194, opened 2018-03-26
Move Style wiki page to whatwg.org

GloverDonovan:
This commit adds the Style wiki page to whatwg.org. Since `style` is already a directory, I opted to go with `style-guide` instead. (I also believe that "Style Guide" is the better title here.)

This is a direct copy of the style page. I think it's best to improve the page with separate issues / pull requests. Fixed a typo for "possessives". Besides that, nothing else should have changed.

Ref #162

### DIFF
--- a/whatwg.org/style-guide
+++ b/whatwg.org/style-guide
@@ -1,0 +1,91 @@
+<!DOCTYPE HTML>
+<html lang="en">
+<meta charset="utf-8">
+<title>Style Guide — WHATWG</title>
+<meta name="viewport" content="width=device-width, initial-scale=1.0">
+<meta name="theme-color" content="#3A7908">
+<link rel="icon" href="https://resources.whatwg.org/logo.svg">
+<link rel="stylesheet" href="/style/shared.css">
+<link rel="stylesheet" href="/style/subpages.css">
+
+<header>
+ <hgroup>
+  <h1>
+   <a href="/">
+    <img id="main-logo" src="https://resources.whatwg.org/logo.svg" alt="">
+    WHATWG
+   </a>
+  </h1>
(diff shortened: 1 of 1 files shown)


annevk:
This looks good to me. I guess we should also start linking this from more places.

GloverDonovan:
I think it's best to link from https://participate.whatwg.org.

https://whatwg.org/policies may also be ideal.

domenic:
I think the whatwg/meta contributing.md is the place to link from. Or maybe we should just move this document to whatwg/meta?

annevk:
That'd also work for me, although I'm somewhat hesitant to ask @GloverDonovan to convert it again. @GloverDonovan thoughts?

GloverDonovan:
I don't think this matters too much. Let's just pick a home for the style guide and stick with it.

From my perspective, the style guide is more about WHATWG itself than GitHub. Putting it on whatwg.org makes referencing the link short and memorable.

annevk:
Yeah, I like putting things on our website better as well. Let's do that and then reference it from whatwg/meta.

domenic:
I don't think this is about the organization, but instead about how to contribute on GitHub. And making it a website makes it harder to edit.

If we think this should be on the website, I'd like to get some idea of why we have any documents in whatwg/meta at all, and perhaps get all of those converted too.

annevk:
I tend to see it as things that are tied to GitHub might as well be here, but things that are more generic would be nice to put on the website, so they can outlive GitHub (hopefully not, but who knows).

annevk:
Sorry for the delay here.

GloverDonovan:
No problem. Thanks for merging this!

PR #196, opened 2018-04-13
Remove analytics

annevk:
It turns out that the one reason we kept it for we didn't end up using it for.

### DIFF
--- a/whatwg.org/index.html
+++ b/whatwg.org/index.html
@@ -14,7 +14,7 @@ body {
  min-height: 100vh;
 }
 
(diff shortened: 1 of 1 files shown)


domenic:
For historical archeologists, "the one reason" we kept it was that we thought it would be nice to look at browser usage share and use that to guide decisions on what sort of JS/CSS/HTML features we could use when building the website. But, we never actually went and looked at the analytics dashboard, so this idea was just theoretical, and not worth maintaining analytics for.

PR #198, opened 2018-05-03
Copy working-mode and code-of-conduct from whatwg/sg

domenic:
This removes the two-sources-of-truth problem, and keeps them updated. It doesn't solve #133, which is about ensuring that updates get automatically published when whatwg/sg changes.


deploy.sh
```diff
@@ -8,6 +8,8 @@ git clone --depth=1 https://github.com/whatwg/sg sg
 ./convert-policy.py sg/SG\ Agreement.md policy-template.html sg/policy-link-mapping.txt  > whatwg.org/sg-agreement
 ./convert-policy.py sg/Principles.md policy-template.html sg/policy-link-mapping.txt  > whatwg.org/principles
 ./convert-policy.py sg/IPR\ Policy.md policy-template.html sg/policy-link-mapping.txt  > whatwg.org/ipr-policy
+cp sg/code-of-conduct > whatwg.org/code-of-conduct
+cp sg/working-mode > whatwg.org/working-mode
```
annevk: cp doesn't use > I think. I suspect that's why the build script is complaining.

domenic: Thanks, fixed

PR #200, opened 2018-05-09
Refactor CSS to support review drafts

domenic:
Part of https://github.com/whatwg/sg/issues/70.

The dfn.js change is a tagalong; I could split that out if people want.

To deploy this, we'd need to land this and https://github.com/tabatkins/bikeshed/pull/1239 at around the same time, then trigger Travis rebuilds of every spec.


domenic:
Oops, good catch, will fix and force push. And yeah, that's the idea. Any thoughts on a better name? Maybe "standard-or-review-draft.css"? :-/

annevk:
Perhaps `spec.css` after the URL we use?

domenic:
Fixed

Issue #197, opened 2018-04-28
Review Drafts automation


annevk:
I looked into this a bit and one problem is that `copy_extra_files` and `run_post_build_step` only really work on Travis and not for a local deploy. Those should be part of a Review Draft however. And we can't really check in the source and let Travis generate the output since we want that to be done once and commit the output.

So we could try to read `.travis.yml` locally (perhaps not that hard, even in bash?) or revamp the way we do this.

annevk:
Also: `REVIEW_DIR="review-drafts/$(date +'%Y-%m')"` is what we want I think. Yields `review-drafts/2018-04` if run today.

domenic:
Hmm it'd be better to check in the source text and let Travis generate the output if possible... That is, review-drafts/2018-05.bs gets checked in with the source as of that moment in time.

We'd need some way to make it so that it doesn't get re-built every time, but only each time we add a new review draft or update some minor text inside the review draft. That seems pretty doable; we'd be able to use git to see if the file has changed.

domenic:
> as we might change the tooling around and I wouldn't want to have to support decade-old source files forever.

Agreed.

> How would you see Travis generating the output? It generates it and then commits it on the PR branch?

I was thinking just the same way as we do everything else: it generates it then uploads it to the server.

The tricky part is just making it not do that if we haven't changed the source. But it seems doable.

> We'd also need to ensure that further changes don't rejigger the date (if we really wanted to change the date we'd open a new PR).

Bikeshed has a Date metadata that you can set explicitly, thankfully.

annevk:
`git diff --name-status $TRAVIS_COMMIT_RANGE` seemed to be what we need to know the difference between added, modified, and not touched. However, `$TRAVIS_COMMIT_RANGE` is reportedly problematic: https://github.com/travis-ci/travis-ci/issues/2668.

https://developer.github.com/v3/pulls/#list-pull-requests-files might therefore be a better fit. We could process such a response in bash with something akin to https://github.com/nest/nest-simulator/pull/75/files. https://stedolan.github.io/jq/manual/ has the manual for `jq`.

Apart from that we need some client-side instruction that copies the source and puts it in the correct directory. This probably fits in the `Makefile` and we'd just copy-and-paste.

domenic:
[...]

PR #201, opened 2018-05-09
WIP review draft deploy

domenic:
See #197 and https://github.com/whatwg/sg/issues/70.

This doesn't yet prevent re-building and deploying; discussions ongoing in #197.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -112,15 +112,20 @@ run_post_build_step "$WEB_ROOT"
 echo "Living standard output to $WEB_ROOT"
 echo ""
 
-header "Starting review drafts..."
+header "Starting review drafts (if applicable)..."
 echo "Note: review drafts must be added or changed in a single commit on master"
 for REVIEW_DRAFT in "$REVIEW_DRAFTS_DIR"/*.bs; do
-    [ -e "$REVIEW_DRAFT" ] || continue # http://mywiki.wooledge.org/BashPitfalls#line-80
+    # http://mywiki.wooledge.org/BashPitfalls#line-80
+    if [[ ! -e "$REVIEW_DRAFT" ]]; then
+        continue
+    fi
+
     if [[ "$TRAVIS_PULL_REQUEST" == "true" ]]; then
         CHANGED_FILES=$(git diff --name-only master..HEAD)
     else
         CHANGED_FILES=$(git diff --name-only HEAD~1)
     fi
+
     for CHANGED in $CHANGED_FILES; do # Omit quotes around variable to split on whitespace
         if [[ "$REVIEW_DRAFT" != "$CHANGED" ]]; then
             continue
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -111,6 +111,21 @@ run_post_build_step "$WEB_ROOT"
 echo "Living standard output to $WEB_ROOT"
 echo ""
 
+header "Starting review drafts..."
+for f in "$REVIEW_DRAFTS_DIR"/*.bs; do
+ [ -e "$f" ] || continue # http://mywiki.wooledge.org/BashPitfalls#line-80
+
+ BASENAME=$(basename "$f" .bs)
+ DRAFT_DIR="$WEB_ROOT/review-drafts/$BASENAME"
+ mkdir -p "$DRAFT_DIR"
+ curlbikeshed -F md-Status="RD" \
+              > "$DRAFT_DIR/index.html"
+ copy_extra_files "$DRAFT_DIR"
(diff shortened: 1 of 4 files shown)


resources.whatwg.org/build/deploy.sh
```diff
@@ -111,6 +112,32 @@ run_post_build_step "$WEB_ROOT"
 echo "Living standard output to $WEB_ROOT"
 echo ""
 
+header "Starting review drafts..."
+echo "Note: review drafts must be added or changed in a single commit on master"
+for f in "$REVIEW_DRAFTS_DIR"/*.bs; do
+    [ -e "$f" ] || continue # http://mywiki.wooledge.org/BashPitfalls#line-80
+    if [[ "$TRAVIS_PULL_REQUEST" == "true" ]]; then
+        CHANGED_FILES=$(git diff --name-only master..HEAD)
```
annevk: Since we put the DATE directly in the draft I don't think we need to distinguish between added and changed anymore. Hence we can simply use --name-only instead of --name-status.

I used master..HEAD here so a PR will consider all commits. We'll just have to ensure to land as a single commit on master.

resources.whatwg.org/build/deploy.sh
```diff
@@ -111,6 +112,32 @@ run_post_build_step "$WEB_ROOT"
...
+    else
+        CHANGED_FILES=$(git diff --name-only HEAD~1)
+    fi
+    for change in "$CHANGED_FILES"; do
+        if [[ f != change ]]; then
```
annevk: Or should these variables be quoted again here? It's also a little unclear to me when to user lowercase and when to use uppercase for variables.

resources.whatwg.org/build/deploy.sh
```diff
@@ -111,6 +112,32 @@ run_post_build_step "$WEB_ROOT"
...
+            continue
+        fi
+        echo ""
+        BASENAME=$(basename "$f" .bs)
+        DRAFT_DIR="$WEB_ROOT/$REVIEW_DRAFTS_DIR/$BASENAME"
+        mkdir -p "$DRAFT_DIR"
+        curlbikeshed -F md-Status="RD" \
```
annevk: Should this be whatwg/RD?

resources.whatwg.org/build/deploy.sh
domenic: Nice!

resources.whatwg.org/build/deploy.sh
domenic: Nah, since we're already in the WHATWG org, just the status suffices.

resources.whatwg.org/build/deploy.sh
```diff
@@ -111,6 +112,32 @@ run_post_build_step "$WEB_ROOT"
...
+header "Starting review drafts..."
+echo "Note: review drafts must be added or changed in a single commit on master"
+for f in "$REVIEW_DRAFTS_DIR"/*.bs; do
+    [ -e "$f" ] || continue # http://mywiki.wooledge.org/BashPitfalls#line-80
+    if [[ "$TRAVIS_PULL_REQUEST" == "true" ]]; then
+        CHANGED_FILES=$(git diff --name-only master..HEAD)
+    else
+        CHANGED_FILES=$(git diff --name-only HEAD~1)
+    fi
+    for change in $CHANGED_FILES; do # Omit quotes around variable to split on whitespace
```
domenic: FWIW we should probably stick to uppercase everywhere, I was just copying and pasting from places that did lowercase.

resources.whatwg.org/build/review.sh
```diff
@@ -0,0 +1,16 @@
...
+
+# This script is maintained at
+# https://github.com/whatwg/whatwg.org/tree/master/resources.whatwg.org/build.
+# See README.md there for documentation.
+
+mkdir -p "review-drafts"
+INPUT_FILE=$(find . -maxdepth 1 -name "*.bs" -print -quit)
+REVIEW_DRAFT="review-drafts/$(date +'%Y-%m').bs"
+# The very unusual way of creating a newline here is done to keep macOS happy
+sed 's/^Group: WHATWG$/&'''$'\n'"Date: $(date +'%Y-%m-%d')/g" < "$INPUT_FILE" > "$REVIEW_DRAFT"
```
foolip: I tried running this locally and got <code>sed: -e expression #1, char 19: unterminated `s' command</code> which is probably about this line. Using sed (GNU sed) 4.4.

annevk: Hmm, I suspect the various versions of sed are incompatible in their syntaxes. What matters most is that this works on Travis, which we haven't really tried yet. But it does pass shellcheck apparently.

annevk: I think `'''` also does not work on macOS. `\'` did work.

PR #202, opened 2018-05-09
Add service workers to whatwg.org and {spec,idea}.whatwg.org

### LINKED ISSUE #152: Make whatwg.org and spec.whatwg.org available offline via a service worker
I don't think going "full PWA" with a manifest is worthwhile since currently only Chrome uses the manifest and it requires hiding the address bar to get any benefits. (See previous discussions in https://github.com/whatwg/streams/pull/637.)

But a service worker at least would be nice.

domenic:
Closes #152.

We'll also need to update working-mode and code-of-conduct in the SG repo.


idea.whatwg.org/service-worker.js
```diff
@@ -0,0 +1,16 @@
...
+  "https://whatwg.org/style/subpages.css",
+  "https://resources.whatwg.org/logo.svg"
+];
+
+importScripts("https://resources.whatwg.org/website-service-worker.js");
+
+<script>
+"use strict";
+navigator.serviceWorker.register("/service-worker.js");
+</script>
```
annevk: This shouldn't be here it seems?

domenic: Thanks, fixed and re-pushed

PR #205, opened 2018-05-14
Copy policies from whatwg/sg

annevk:
See https://github.com/whatwg/sg/pull/77.


annevk:
As per that issue I'd prefer merging myself to deal with any fallout.

PR #208, opened 2018-05-30
Fix some review draft creation scripts

domenic:


domenic:
PR renamed and added another fix that I found

PR #214, opened 2018-07-11
Fix syntax highlighting for standards

domenic:
We accidentally broke it in https://github.com/tabatkins/bikeshed/commit/7a022f4ed5d215ee6aab86547be6126ba9c8135a, thinking the block added in #209 would suffice, but due to https://github.com/tabatkins/bikeshed/issues/1313 it does not work yet.

While waiting for https://github.com/tabatkins/bikeshed/issues/1313 to be fixed, include these styles, to un-break syntax highlighting.


tabatkins:
Sorry for the delay, Bikeshed now uses the same highlighting elements as Highlighter.

PR #219, opened 2018-07-25
Website service worker fixups

domenic:
968af02faaa8eb00673ddbd93411d2895fb772e5 accidentally included some URLs that don't exist in the to-cache list.

Oops.

### DIFF
--- a/whatwg.org/service-worker.js
+++ b/whatwg.org/service-worker.js
@@ -1,6 +1,5 @@
 "use strict";
 
-self.cacheKey = "v2";
 self.toCache = [
   "/img/arrow.svg",
   "/img/bird-webpage.svg",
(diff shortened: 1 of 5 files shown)


domenic:
Wait, there was more weirdness...

whatwg.org/service-worker.js
```diff
@@ -1,10 +1,7 @@
 "use strict";
 
-self.cacheKey = "v1";
+self.cacheKey = "v2";
```
annevk: It seems you could remove this given the default, but maybe it's better if there's no default and each of these service-worker.js files have it explicitly so we don't forget about it when we update those files?

domenic: I think upon reflection the better thing to do is concatenate them. We want to have the cache refresh if the base service worker list of resources/fetching logic changes, or if the individual list changes. Let me do that instead.

PR #223, opened 2018-08-21
Force 0 exit status when running bikeshed for errors/warnings

annevk:
See https://github.com/whatwg/quirks/pull/30#issuecomment-414734994.


annevk:
That seems better than not telling those making the PR what went wrong. I'm not sure why we'd optimize computers over humans.

domenic:
The point is that you should bail if errors are detected and printed, instead of making the user wait for a second server round trip before the script exits.

annevk:
Feel free to push a fixup, I don't really know what that would look like.

alauq (commented):
[]()

PR #224, opened 2018-08-21
Small fix to service worker error handling

domenic:
Before this, navigating to a page that gives a non-OK status for a living standard (e.g. https://streams.spec.whatwg.org/asdf) would cause the code to go down a path of trying to get the result from the cache. Since the cache would contain no entries for such pages, this would end up calling fetchEvent.respondWith(null). The result is a network error, giving a scary error page in various browers (e.g. "Corrupted Content Error" in Firefox).

After this page, the non-OK status only impacts the waitUntil() call, as intended.


annevk:
Last page in the commit message should be change.

PR #225, opened 2018-08-29
Abort deploy when Bikeshed finds a warning

annevk:
This addresses part of #176.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -72,6 +72,10 @@ curlbikeshed() {
     curlretry https://api.csswg.org/bikeshed/ -F file=@"$INPUT_FILE" "$@"
 }
 
+curlbikesheddie() {
+    curlbikeshed -F die-on=warning "$@"
+}
+
 header "Linting the source:"
 MATCHES=$(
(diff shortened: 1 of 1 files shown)


annevk:
Other than https://github.com/tabatkins/bikeshed/issues/1338#issuecomment-416889345 I think we should be able to fix all outstanding warnings.

The problem with the current setup is that I had at one point fixed all warnings and then they came back, so I think we should just land this and take the hit.

@miketaylr I noticed Compat has a fair number of warnings so this might affect you a bit as well. I'd be happy to help out with PRs if you get stuck.

annevk:
That would not be possible. You'd basically have to figure out how to resolve the conflict first.

Note that we already accepted this in how we do things. Yesterday a change to Bikeshed was made that prevents the Encoding Standard (and possibly all standards) from building: https://github.com/tabatkins/bikeshed/pull/1340.

foolip:
Yes, that change did block deploy of Fullscreen too. I guess if we get an issue that blocks us we will temporarily revert this change?

annevk:
Yeah or we'll just hold the PR a little until it's fixed. I guess it depends on the severity. @domenic thoughts?

foolip:
> hold the PR

Unfortunately we'll notice the problem once the PR is merged and not before, at least that was the case with Fullscreen.

annevk:
Yeah, if Bikeshed changes that would indeed happen (or existing PRs are not refreshed after we land this). That's likely very hard to avoid other than force pushing the PR before landing it on master shortly after (if it was successful).

domenic:
I'm fine with Bikeshed having more ways to break our builds. You can imagine better systems, but we're already committed to that kind of workflow, and this just expands the scope a bit.

annevk:
I simplified this per https://github.com/whatwg/whatwg.org/issues/176#issuecomment-416896771. If this looks good I'll merge it.

PR #226, opened 2018-08-30
Output Bikeshed errors if they occur

domenic:
Closes #223.

This works best only after #225 is merged, as otherwise it loses warning information (because it only shows the output if the web service gave a non-200). Happy to rebase it afterward.

This reduces the number of HTTP requests by one in the happy case, since we don't need a dedicated warning/error check pass anymore. It also avoids the problem of #223 and only does one HTTP request in the unhappy case before bailing.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -65,11 +65,23 @@ header() {
 }
 
 curlretry() {
-    curl --fail --retry 2 "$@"
+    curl --retry 2 "$@"
 }
 
 curlbikeshed() {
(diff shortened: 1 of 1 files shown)


annevk:
I don't really understand how this works. Don't $0 and $1 refer to the arguments passed to the function?

domenic:
Ah oops, typo. It should be $1 in both cases. Function arguments start from $1.

domenic:
This did not work as expected. The interaction with `die-on` was not what I hoped; if it dies on a warning, the Bikeshed web service still outputs the HTML.

So I think we'll need the extra HTTP request after all. Unless @plinss is up for adding a mode where `die-on=warning` causes the output to be the warning messages, instead of the HTML.

In the meantime I'll revert.

PR #232, opened 2018-10-18
Fix Markdown compilation deployment

domenic:
A recent update to the Python Markdown module broke our build, as the py-gfm module we use isn't compatible with markdown 3.x. This pins the versions of our Python modules to avoid such breakage.


alphachart (commented):
.

PR #235, opened 2018-10-26
Remove one over-shared spec style

domenic:
Follows https://github.com/whatwg/html/pull/4130. This style only applies to the dev edition, and should not be shared.


domenic:
I guess I should have mentioned how to know this is wrong: if you go to https://infra.spec.whatwg.org/#conformance, the h2's self-link is positioned off-screen. Whereas if you go to https://html.spec.whatwg.org/dev/infrastructure.html#infrastructure you can see the -4em is correct.

PR #236, opened 2018-11-06
Fix "good first bugs" link in FAQ (bug->issue)

foolip:


foolip:
Thanks @marcoscaceres for pointing this out on IRC.

PR #237, opened 2019-01-18
Make review.sh deal with git

annevk:


resources.whatwg.org/build/review.sh
```diff
@@ -1,10 +1,14 @@
 #!/bin/bash
-set -o errexit
```
domenic: Why remove this? If it's for diff, then I think a better solution would be to add `|| 0` to the end of that line

annevk: When I try that on whatwg/html's equivalent I get `./review-draft.sh: line 26: 0: command not found`.

sideshowbarker: I reckon Domenic meant `|| exit 0`

domenic: It looks like `true` is what I meant. I will add suggested changes here.

resources.whatwg.org/build/review.sh
```diff
@@ -15,4 +19,7 @@ sed 's/^Group: WHATWG$/&\
 '"Date: $(date +'%Y-%m-%d')/g" < "$INPUT_FILE" > "$REVIEW_DRAFT"
 echo "Created Review Draft at $REVIEW_DRAFT"
```
domenic:
```suggestion
echo ""
echo "Created Review Draft at $REVIEW_DRAFT"
```

resources.whatwg.org/build/review.sh
```diff
@@ -15,4 +19,7 @@ sed 's/^Group: WHATWG$/&\
...
 echo "Please verify that only one line changed relative to $INPUT_FILE:"
-diff -up "$INPUT_FILE" "$REVIEW_DRAFT" || exit 0
+diff -up "$INPUT_FILE" "$REVIEW_DRAFT"
```
domenic:
```suggestion
echo ""
diff -up "$INPUT_FILE" "$REVIEW_DRAFT" || true
echo ""
```

annevk:
Thanks, addressed! I also have the changes for whatwg/html locally (and tested). I guess if this is all okay I'm going to do a new PR for that one (and the other drafts).

PR #240, opened 2019-03-06
Working Mode is in Markdown now

annevk:
See https://github.com/whatwg/sg/pull/88. Needs to be merged afterwards. And also requires #295 and rebasing.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,8 +9,6 @@ whatwg.org/sg-policy
 whatwg.org/code-of-conduct
 whatwg.org/principles
 whatwg.org/policies
-
-# Not yet covered by the above, but imported from the same repository
 whatwg.org/working-mode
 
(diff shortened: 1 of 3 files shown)


annevk:
See https://github.com/whatwg/sg/pull/88#issuecomment-595805606 and https://github.com/whatwg/whatwg.org/issues/291 for context of the recent changes.

foolip:
I've compared the built output with these changes and it looks like an `<hr>` is gone from all the converted files. Is that known/intended?

foolip:
Yes, that's very much intentional, I see it in the code now :)

foolip:
"Needs to be merged afterwards." left in commit message because I rebased. Oh well.

annevk:
Thanks @foolip!

PR #243, opened 2019-03-15
Modernize file-issue.js and make it run less

annevk:
This change:

* Requires the user to make a selection so we get a useful target node.
* Avoids running the script for each mouseup/keydown.
* Uses const/let throughout.

### DIFF
--- a/resources.whatwg.org/file-issue.js
+++ b/resources.whatwg.org/file-issue.js
@@ -10,43 +10,44 @@
 
 (function () {
   'use strict';
-  var thisScript = document.currentScript;
 
-  var originalFilingUrl = getOriginalFilingUrl();
-  var titlePrefix = '';
-  var queryParamIndex = originalFilingUrl.indexOf('?title=');
+  let originalFilingUrl = getOriginalFilingUrl(),
+      titlePrefix = '';
+  const queryParamIndex = originalFilingUrl.indexOf('?title=');
   if (queryParamIndex != -1) {
     titlePrefix = decodeURIComponent(originalFilingUrl.substr(queryParamIndex + '?title='.length));
     originalFilingUrl = originalFilingUrl.substr(0, queryParamIndex);
   }
 
-  var specUrl = getSpecUrl();
+  const specUrl = getSpecUrl();
 
-  var fileLink = document.createElement('a');
+  const fileLink = document.createElement('a');
   fileLink.href = originalFilingUrl;
   fileLink.accessKey = '1';
   fileLink.className = 'selected-text-file-an-issue';
   fileLink.textContent = 'File an issue about the selected text';
 
+  fileLink.onclick = e => {
+    const selection = window.getSelection();
+    if (selection.toString() === "") {
+      alert("Please select some text first.");
+      e.preventDefault();
+    }
+    fileLink.href = getFilingUrl(originalFilingUrl, selection);
+  };
+
   document.body.insertBefore(fileLink, document.body.firstChild);
 
-  window.addEventListener('mouseup', handleInteraction);
-  window.addEventListener('keydown', handleInteraction);
-
-  function handleInteraction(event) {
-    if (event.target === fileLink) {
-      return;
-    }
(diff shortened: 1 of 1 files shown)


resources.whatwg.org/file-issue.js
```diff
@@ -10,43 +10,44 @@
 
 (function () {
   'use strict';
-  var thisScript = document.currentScript;
 
-  var originalFilingUrl = getOriginalFilingUrl();
-  var titlePrefix = '';
-  var queryParamIndex = originalFilingUrl.indexOf('?title=');
+  let originalFilingUrl = getOriginalFilingUrl(),
+      titlePrefix = '';
```
domenic: Style-wise I am not comfortable maintaining scripts that use comma-separated variable declarations; it's too easy to miss the comma and then end up declaring a global variable.

resources.whatwg.org/file-issue.js
```diff
@@ -10,43 +10,44 @@
...
-  window.addEventListener('keydown', handleInteraction);
+  document.body.insertBefore(fileLink, document.body.firstChild);
 
-  function handleInteraction(event) {
-    if (event.target === fileLink) {
-      return;
-    }
-    fileLink.href = getFilingUrl(originalFilingUrl, window.getSelection(), event.target);
-  }
+  //window.addEventListener('mouseup', handleInteraction);
```
domenic: Let's delete code, not comment it out.

resources.whatwg.org/file-issue.js
```diff
@@ -10,43 +10,44 @@
...
+    if (selection.toString() === "") {
+      alert("Please select some text first.");
+      e.preventDefault();
+    }
+    fileLink.href = getFilingUrl(originalFilingUrl, selection);
+  };
 
-  window.addEventListener('mouseup', handleInteraction);
-  window.addEventListener('keydown', handleInteraction);
+  document.body.insertBefore(fileLink, document.body.firstChild);
```
domenic: You can use `document.body.prepend()` if we're going to assume modern browsers.

resources.whatwg.org/file-issue.js
```diff
@@ -10,43 +10,44 @@
...
   fileLink.href = originalFilingUrl;
   fileLink.accessKey = '1';
   fileLink.className = 'selected-text-file-an-issue';
   fileLink.textContent = 'File an issue about the selected text';
 
-  document.body.insertBefore(fileLink, document.body.firstChild);
+  fileLink.onclick = e => {
+    const selection = window.getSelection();
+    if (selection.toString() === "") {
+      alert("Please select some text first.");
```
domenic: I'm not a big fan of using alert in new code, even if it is expedient in this way. Would failing silently or with a console message be acceptable? Or, just letting people file issues even without text selected, like the current UI does?

resources.whatwg.org/file-issue.js
```diff
@@ -55,22 +56,22 @@
...
     return originalFilingUrl + '?title=' + encodeURIComponent(bugData.title) + '&body=' +
            encodeURIComponent(bugData.body);
   }
 
-  function getBugData(selection, startNode) {
-    var selectionText = selection.toString();
-    var url = getUrlToReport(selection, startNode);
+  function getBugData(selection) {
+    const selectionText = selection.toString(),
+          url = getUrlToReport(selection);
```
domenic: Comma again here

resources.whatwg.org/file-issue.js
annevk: That would work too.

annevk:
I only tested this on desktop. I don't really have a good setup to test it elsewhere.

zcorpan:
Did you try middle-clicking to open in a tab?

We could consider using target=_blank

annevk:
I did, it works fine in Firefox at least.

zcorpan:
OK good. I don't have a mouse with a third button to test, but recall there were issues with this before.

https://bugs.chromium.org/p/chromium/issues/detail?id=255

Looks like using `auxclick` as well may be necessary?

PR #245, opened 2019-03-24
Suggest the WICG as a way to propose new features.

jyasskin:
The WICG can provide initial discussion of a feature and dispatch between the many
available repositories on which one could file the issue.

Inspired by https://twitter.com/troglotit/status/1109577257738227712.

May help with @gsnedders' https://github.com/whatwg/participate.whatwg.org/issues/37 but doesn't fix it.

### DIFF
--- a/whatwg.org/faq
+++ b/whatwg.org/faq
@@ -179,11 +179,11 @@ in-progress removal, which may or may not be successful, is the
  use case. For a good example of how to do this, see
  <a href="http://lists.w3.org/Archives/Public/public-webapps/2012JulSep/0835.html">this email</a>.
 
- <li><p>Get more people involved. Open a new issue in
- <a href="https://github.com/whatwg/html/issues">whatwg/html on GitHub</a> that describes the use
- cases and their requirements. Ask fellow web developers about their opinions (but remind them of
- step 1 above). Adjust the list of use cases and requirements as appropriate. Say which use cases
- are important and which are just nice to have.
+ <li><p>Get more people involved. Start a thread in <a href="https://discourse.wicg.io/">the
+ WICG</a> or open a new issue in <a href="https://github.com/whatwg/html/issues">whatwg/html on
+ GitHub</a> that describes the use cases and their requirements. Ask fellow web developers about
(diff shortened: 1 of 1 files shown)


domenic:
It makes more sense to me to fix this sentence to generalize whatwg/html to "the standard's issue tracker". And maybe say "if you're not sure what standard your idea impacts, use whatwg/meta. If your idea is outside the scope of WHATWG standards entirely, use WICG."

jyasskin:
I like "the standard's issue tracker" and will make that change later today. I think it's often difficult for people outside our community to know which standard their thing is supposed to affect, so I'll try to word something that suggests that either whatwg/meta or the WICG can be used to dispatch problem statements to particular specifications?

gsnedders:
@domenic: "If your idea is outside the scope of WHATWG standards entirely, use WICG." — That's not immediately obvious if you don't spend enough time around here, IMO.

jyasskin:
Thinking about this more, it doesn't make sense to try to guess the standard the solution is eventually going to land in, at a point where we're insisting that people should only have a problem statement and not a proposed solution.

domenic:
I think in a lot of cases people do know what standard something would land in. E.g. if they're suggesting a new way of fetching things, they might know the fetch standard is the place to be. Or if they're talking about addressing a deficiency with a HTML tag, they might know HTML is the place to be. If they're concerned about a lack of capabilities of notifications, it makes sense to file the issue on whatwg/notifications. Etc.

cwilso:
I'd suggest that if people aren't spending enough time around WHATWG to know that their proposal belongs there, it's not a failure mode to suggest going to WICG - they'd be steered back in the end if WHATWG is more appropriate.  WICG is intended to be a reasonable home for incubation of new features for anything, whether they end up in W3C or not.

I agree with @domenic that when people know where a feature would land, proposing it there (in whatwg/notifications or wherever) is probably the shortest path to success, of course.

jyasskin:
Done, thanks.

domenic:
I feel like this has been overtaken by https://whatwg.org/stages ; WDYT?

annevk:
Although maybe we should update the FAQ to account for that.

PR #248, opened 2019-04-28
Uniformize spec paragraph styles

### LINKED ISSUE #187: [standard.css] Make padding consistent

domenic:
In particular, this uniformizes .domintro, .note, .warning, .example, .XXX, and .critical, in the following ways:

* Updates all of .domintro, .note, .warning, and .example to have the same style as the current .domintro for their decoration, with no tilt
* Updates all to the look the same for `<p class="thing">...</p>` and `<div class="thing"><p>...</p></div>`
* Updates all of them to have uniform padding and margin, so that e.g. the text aligns on the left, and the borders align on the left and right
* Improves the style for inline `<span>` versions of .note, .warning, and .XXX
* Removes the .big-issue synonym for .XXX
* Removes many redundant-in-practice rules that have accreted over time and serve only to create inconsistencies
* Makes it clearer what styles are shared, by explicitly sharing them, instead of repeating them
* Also de-italicizes `<cite>` inside italic contexts like .note or .warning (similar to `<em>`, `<i>`, and `<var>`)

Closes #187. Closes #189.

---

You can see a before/after comparison at:

- https://condescending-franklin-fe08bc.netlify.com/test-spec.html
- https://condescending-franklin-fe08bc.netlify.com/test-spec-new.html

If I were to tweak something further, I would add a background color to .warning; if folks have suggestions that'd be helpful.

### DIFF
--- a/resources.whatwg.org/standard.css
+++ b/resources.whatwg.org/standard.css
@@ -172,39 +172,141 @@ dl.triple dt:after { content: ':'; }
 dl.triple dd:after { content: '\A'; white-space: pre; }
 tr.rare { background: #EEEEEE; color: #333333; }
 
-dl.domintro { position: relative; color: green; background: #DDFFDD; margin: 2.5em 0 2em 0; padding: 0.5em 1em 0.5em 2em; }
-hr + dl.domintro, div.impl + dl.domintro, hr + div.status + dl.domintro, div.impl + div.status + dl.domintro { margin-top: 2.5em; margin-bottom: 1.5em; }
-dl.domintro dt, dl.domintro dt * { color: black; }
-dl.domintro dd { margin: 0.5em 0 1em 2em; padding: 0; }
-dl.domintro dd p { margin: 0.5em 0; }
-dl.domintro::before { background: green; color: white; padding: 0.15em 0.25em; font-style: normal; position: absolute; top: -0.8em; left: -0.8em; }
+/*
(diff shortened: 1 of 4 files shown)


resources.whatwg.org/standard.css
```diff
@@ -172,40 +172,142 @@ dl.triple dt:after { content: ':'; }
...
+  border: solid red;
+}
+
+.critical {
+  background: #FFFFCC;
+  border: double thick red;
+}
+
+
+
```
annevk: Why the triple newline?

domenic: That was my attempt at separating out this "section" of CSS. Other ideas welcome. I thought

```css
/*
 * end .domintro, .note, .warning, .example, .XXX, .critical
 */
```

was too heavy, but that's another idea. Or we could just have a normal newline.

annevk: Current pattern is a normal newline, see e.g. self-links. Not sure if it's great though.

Issue #246, opened 2019-04-08
Lint for tabs


annevk:
Apparently in https://github.com/whatwg/html/pull/4455 tabs got into a source document somehow. That'd be easy to prevent via linting.

annevk:
There's potentially a whole blocklist here I suppose, but only addressing the known issues for now seems fine.

domenic:
There are a few things to figure out here:

- build.sh in this repository doesn't seem to fail the build on lint errors. It just prints them. That's weird.
- build.sh in this repository inlines the lint script. This is separate from https://github.com/whatwg/html-build/blob/master/lint.sh. Should we merge them?
  - Pro: shared linting heuristics
  - Con: makes html-build a bit more complicated and slow, since it needs to download the lint script on every use. (Or we could download it only on first use, but then we'll have out of sync local lint scripts.) 
  - Con: slight duplication of effort and complications, e.g. html-build needs manual RFC 2119 linting but Bikeshed does that for us here; html-build has an extra lint line for "Possible copypasta" that does not appear here.

I'll update the script here but we should either leave this open or open a new issue for html-build, especially since the original issue was about a html PR.

PR #252, opened 2019-05-11
Lint for tabs, and fail on lint errors

domenic:
Part of #246, but does not handle HTML.

https://askubuntu.com/questions/53071/how-to-grep-for-tabs-without-using-literal-tabs-and-why-does-t-not-work is why I used `-P` here.


domenic:
This closes https://github.com/whatwg/whatwg.org/issues/176; we should amend the commit message when merging to reflect that.

resources.whatwg.org/build/deploy.sh
```diff
@@ -93,9 +93,11 @@ MATCHES=$(
   perl -ne '$/ = "\n\n"; print "$_" if (/\ban\s+(<[^>]*>)*(?!(L\b|http|https|href|hgroup|rb|rp|rt|rtc|li|xml|svg|svgmatrix|hour|hr|xhtml|xslt|xbl|nntp|mpeg|m[ions]|mtext|merror|h[1-6]|xmlns|xpath|s|x|sgml|huang|srgb|rgb|rsa|rst_stream|only|option|optgroup)\b|html)[b-df-hj-np-tv-z]/si or /\b(?<![<\/;])a\s+(?!<!--grammar-check-override-->)(<[^>]*>)*(?!&gt|one)(?:(L\b|http|https|href|hgroup|rt|rp|li|xml|svg|svgmatrix|hour|hr|xhtml|xslt|xbl|nntp|mpeg|m[ions]|mtext|merror|h[1-6]|xmlns|xpath|s|x|sgml|huang|srgb|rsa|only|option|optgroup)\b|html|[aeio])/si)' "$INPUT_FILE" | perl -lpe 'print "\nPossible grammar problem: \"a\" instead of \"an\" or vice versa (to override, use e.g. \"a <!--grammar-check-override-->apple\"):" if $. == 1'
   grep -ni 'and/or' "$INPUT_FILE" | perl -lpe 'print "\nOccurrences of making Ms2ger unhappy and/or annoyed:" if $. == 1'
   grep -niE '\s+$' "$INPUT_FILE" | perl -lpe 'print "\nTrailing whitespace:" if $. == 1'
+  grep -P '\t' "$INPUT_FILE" | perl -lpe 'print "\nTab:" if $. == 1'
```
sideshowbarker:
```suggestion
  grep $'\t' "$INPUT_FILE" | perl -lpe 'print "\nTab:" if $. == 1'
```

sideshowbarker: The `-P` switch is only in GNU grep — it’s not in, for example, the grep that comes in the macOS shell (FreeBSD).

So while we don’t need to worry about that for the case of this particular script — since I guess it’s running in a Linux/GNU shell environment — it’d be better practice to use a more-portable alternative.

The `$'\t'` thing is a Bash-ism but since this script is already explicitly a Bash script — and I guess we have other Bash-isms in it already — then doing `$'\t'` seems preferable.

resources.whatwg.org/build/deploy.sh
```diff
@@ -93,9 +93,11 @@ MATCHES=$(
...
 )
 if [ -n "$MATCHES" ]; then
   echo "$MATCHES"
+  exit 1
```
annevk: We should check all specifications to ensure they don't already fail for a reason we might not be able to fix.

domenic: Failures from https://travis-ci.org/whatwg:

- https://travis-ci.org/whatwg/xhr unicorns-are-teh-awesome, a HTML document (real errors)
- https://travis-ci.org/whatwg/mimesniff an mp3 (script needs fixing)
- https://travis-ci.org/whatwg/compat an RGBA (script needs fixing)

I'll fix the script for mp3/rgba

annevk: unicorns-are-teh-awesome is an intentional misspelling, but I guess there's an opt-out for those now?

annevk: There's not.

domenic: Should we add one?

annevk: I ended up fixing the XHR issue. I'd be okay with not adding anything until there's a more justified need.

domenic: Cool. Does this need anything else before it can be merged, then?

PR #254, opened 2019-05-12
Remove `package-lock.json` file

alrra:
In order to get the dependency updates within the specified semver range, remove the `package-lock.json` file, and configure `npm` to not generate it.

- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -

From https://docs.npmjs.com/files/package-lock.json:

> `package-lock.json` is automatically generated for any operations where `npm` modifies either the node_modules  tree, or `package.json`. It describes the exact tree that was generated, such that subsequent installs are able to generate identical trees, **regardless of intermediate dependency updates**.


.npmrc
```diff
@@ -0,0 +1 @@
+package-lock=false
```
domenic: Why is this necessary?

alrra: @domenic It is not necessary, but my thinking was that it makes things more "consistent" in behavior. 

Let's presume your install the packages locally, without it, `npm` will still generate the `package-lock.json` file that, for example, will lock package `P` to version `x.y.1`. When version `x.y.2` is released, locally (unless  the `package-lock.json` is removed or modified) you will still get version `x.y.1`, while on Travis (because with every build you start from scratch and a new `package-lock.json` is created), you will get `x.y.2`. 

---

Let me know if you want me to remove it.

domenic: I guess that makes sense; thank you!

PR #255, opened 2019-05-12
Conformance check the deployed HTML

### LINKED ISSUE #10: Run HTML checker on resources
Although deploy should only happen on commit to master, we should be able to check the HTML (of "active" resources) all the time.

domenic:
Closes #10.

Unclear whether this will work with all our extension-less files; let's see what Travis does.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -56,7 +56,7 @@ if [[ "$TRAVIS" == "true" ]]; then
     done < <(find whatwg.org -maxdepth 1 -type f ! -name "*.*" ! -name "status-2008-12" -print0)
 
     curl --retry 2 --fail --remote-name --location https://github.com/validator/validator/releases/download/linux/vnu.linux.zip
-    unzip vnu.linux.zip
+    unzip -qq vnu.linux.zip
     ./vnu-runtime-image/bin/vnu --Werror "${TARGETS[@]}"
     echo ""
 fi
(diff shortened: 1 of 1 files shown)


domenic:
OK, I think I figured out the shell scripting part of this. Commits squashed down into one. But, surprise, we have some errors in our HTML! I might be running out of steam for today, so folks should feel free to push more to this branch to help fix them.

annevk:
`n.whatwg.org` should be excluded as those are text files. Same for `whatwg.org/status-2008-12`.

I created patches for the SG-owned files, but I'm not sure if they'll fix the issues completely. Not entirely sure how IPR Policy ended up with stray `</p>` tags.

domenic:
Yeah, the fact that our Markdown processor can produce invalid HTML is not great. I imagine switching to CommonMark might help some, but even then not always, e.g. [as we noted when trying to convert the working mode](https://talk.commonmark.org/t/is-it-possible-to-create-a-paragraph-with-a-class-whose-contents-are-interpreted-as-markdown/3143?u=domenic).

annevk:
I rebased this to see what's left, but the checker can no longer be successfully invoked unfortunately. @sideshowbarker thoughts?

sideshowbarker:
OK, I pushed a change to the branch which makes the build use the Linux binary runtime image of the HTML checker rather than the jar, eliminating the need to rely on the system Java runtime. That should work better for maintenance going forward.

So the HTML checker is now working again on this branch, but Travis is currently back to failing for the different/original reason that the IPR Policy document has stray `</p>` end tags.

annevk:
@sideshowbarker it passes now. Can we make it so that unzip doesn't create a lot of noise in the log?

sideshowbarker:
> Can we make it so that unzip doesn't create a lot of noise in the log?

Yup, added the `-qq` option to the unzip call

annevk:
Great, let's commit it now so we cannot regress it again.

PR #258, opened 2019-06-19
Use --filterfile for HTML checker

sideshowbarker:
This change causes the HTML checker to be invoked with the --filterfile argument if a file named `.htmlcheckerfilter` is found. The filter pattern (for suppressing particular messages from the checker) is then read from the specified file.

Otherwise, without this change, the HTML checker is instead invoked with the --filterpattern argument and the filer pattern is read directly from the CHECKER_FILTER environment variable. Testing in Travis CI indicates doing it that way doesn’t work as expected in some cases (due to shell escaping of the special characters and spaces in the pattern, or something.)


sideshowbarker:
> Could we instead just look for .htmlcheckerfilter or something always?

ah yeah of course that makes more sense. Will switch it to that

> Also can we document this in the README?

Yeah will patch that now too

PR #265, opened 2019-10-22
Update Firefox logo

alrra:
| Old logo | New Logo |
|:---|:---|
|  |


domenic:
It's so cool that you set up the infrastructure so that this requires just changing a version string! Confirmed it's all deployed and showing up in the MDN panels.

PR #269, opened 2019-12-10
Revamp /mailing-list for shutdown of lists.whatwg.org

foolip:


foolip:
This shouldn't be merged yet, but is ready for review. If we have agreement that this is a good way to sunset the lists, the steps would be:
- Email whatwg@whatwg.org and outline this plan
- Configure whatwg@whatwg.org and *@lists.whatwg.org to auto-reply pointing to this updated page
- Update DNS of lists.whatwg.org and deploy https://github.com/whatwg/whatwg.org/pull/269

whatwg.org/mailing-list
```diff
@@ -27,103 +27,41 @@
...
-</div>
+<p>The <a href="https://lists.w3.org/Archives/Public/public-whatwg-archive/">whatwg@whatwg.org
+archives</a> maintained by the W3C are publicly accessible.</p>
+
+<p>Partials archive of whatwg@whatwg.org as well as "help" and "implementors" lists are available
+in the
+<a href="https://web.archive.org/web/20140401085615/http://lists.whatwg.org/listinfo.cgi">Wayback
+Machine</a>.
+
+Given a link to lists.whatwg.org, you can try finding it in the Wayback Machine:
```
annevk:
```suggestion
<p>Given a link to lists.whatwg.org, you can try finding it in the Wayback Machine:</p>

```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,41 @@
...
-
- <p>Quote enough original text or provide an introduction yourself.</p>
-</div>
+<p>The <a href="https://lists.w3.org/Archives/Public/public-whatwg-archive/">whatwg@whatwg.org
+archives</a> maintained by the W3C are publicly accessible.</p>
+
+<p>Partials archive of whatwg@whatwg.org as well as "help" and "implementors" lists are available
+in the
+<a href="https://web.archive.org/web/20140401085615/http://lists.whatwg.org/listinfo.cgi">Wayback
+Machine</a>.
```
annevk:
```suggestion
Machine</a>.</p>
```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,41 @@
...
+
+Given a link to lists.whatwg.org, you can try finding it in the Wayback Machine:
+<form>
+ <input name=listsurl type=text value="http://lists.whatwg.org/listinfo.cgi" size=80>
```
annevk:
```suggestion
 <input name=listsurl type=url value="http://lists.whatwg.org/listinfo.cgi" size=80>
```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,41 @@
...
+ <input type=submit value=Go>
+</form>
+
+<script>
+var form = document.forms[0];
```
annevk:
```suggestion
const form = document.forms[0];
```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,41 @@
...
+var input = form[0];
```
annevk:
```suggestion
const input = form.firstElementChild;
```

whatwg.org/mailing-list
foolip: I tried to write this code in an ancient style to not have to worry about which browsers it would work on :)

domenic:
FWIW it's likely more trouble than its worth, but I think it'd be possible to perform an automatic mapping from lists.whatwg.org URLs to lists.w3.org URLs.

In particular, messages are numbered sequentially in both. However I think the lists.w3.org URLs restart the sequence number every month, whereas the whatwg.org URLs are numbered in an all-time way.

So a mapping could be created by finding the sequence number that begins each month in the lists.whatwg.org archive, and using the following code:

```js
const [month, year, messageNumber] = extractFromURL(listsWHATWGOrgURL);

// This is the hard part and involves a large lookup table
const startingMessageNumber = getStartingMessageNumberForListsWHATWGOrg(month, year);

const newMessageNumber = (messageNumber - startingMessageNumber + 1).toString().padStart(4, "0");

const listsW3COrgURL = `https://lists.w3.org/Archives/Public/public-whatwg-archive/${year}{month}/${newMessageNumber}.html
```

I guess assembling that lookup table could be done with a bit of automation on web.archive.org?

whatwg.org/mailing-list
```diff
@@ -27,103 +27,42 @@
...
-not know what you are referring to:</p>
-
- <pre>No, I think that's a bad idea. It wouldn't be good for the readers, for instance.</pre>
-
- <p>Quote enough original text or provide an introduction yourself.</p>
-</div>
+<p>The <a href="https://lists.w3.org/Archives/Public/public-whatwg-archive/">whatwg@whatwg.org
+archives</a> maintained by the W3C are publicly accessible.</p>
+
+<p>Partials archive of whatwg@whatwg.org as well as "help" and "implementors" lists are available
```
domenic:
```suggestion
<p>Partial archives of whatwg@whatwg.org as well as "help" and "implementors" lists are available
```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,42 @@
...
+Machine</a>.</p>
+
+<p>Given a link to lists.whatwg.org, you can try finding it in the Wayback Machine:</p>
+
+<form>
+ <input name=listsurl type=url value="http://lists.whatwg.org/listinfo.cgi" size=80>
+ <input type=submit value=Go>
+</form>
+
+<script>
```
domenic: Add `"use strict";`?

whatwg.org/mailing-list
domenic: Even IE11 supports const; I think we should do it.

whatwg.org/mailing-list
domenic: Consider wrapping the whole thing in a `(() => {})()` (requires Edge) or `(function () { })()` (supports IE) wrapper so as not to create global variables?

whatwg.org/mailing-list
```diff
@@ -27,103 +27,42 @@
...
+var form = document.forms[0];
+var input = form[0];
+form.addEventListener('submit', function(event) {
+  event.preventDefault();
+  window.location = 'https://web.archive.org/web/20140401085615/' + input.value;
+});
+// If redirected here from lists.whatwg.org, prefill the URL in the form.
+if (location.search.indexOf('?listsurl=') == 0) {
```
domenic:
```suggestion
if (location.search.indexOf('?listsurl=') === 0) {
```

whatwg.org/mailing-list
```diff
@@ -27,103 +27,42 @@
...
+  input.value = location.search.substr(10);
```
domenic:
```suggestion
  input.value = location.search.substring('?listsurl='.length);
```

whatwg.org/mailing-list
foolip: I went with `URLSearchParams` instead.

foolip:
Alright, JS code made as modern as I can rather than as ancient as I can :)

> FWIW it's likely more trouble than its worth, but I think it'd be possible to perform an automatic mapping from lists.whatwg.org URLs to lists.w3.org URLs.

Without also writing the code to verify the mappings exhaustively I wouldn't have very high confidence in them, because the counters presumably depend on time of arrival, and at times when the lists were more active I presume some messages have been processed in different orders by the two mailing list systems.

I think that if someone wants to save as much as possible here, time might be better spent trying to find someone who's been subscribed since the beginning and has saved every email, and try to reconstruct the public archives from that. Also a lot of work, but it would be preserving discussions that are now nowhere to be found in any public archive.

foolip:
> trying to find someone who's been subscribed since the beginning and has saved every email

Sigh, strike that, https://lists.w3.org/Archives/Public/public-whatwg-archive/ goes back to 2004 and one could try to reconstruct pre-2017 of http://lists.whatwg.org/pipermail/whatwg-whatwg.org/ from that.

foolip:
whatwg@whatwg.org and whatwg@lists.whatwg.org now autoreply with this message:

> The WHATWG mailing lists have been shut down.
>
> Discussions now happen on the repositories for the individual standards, and https://github.com/whatwg/meta can be used for issues without a logical home.
>
> See https://whatwg.org/mailing-list for list archives and more information.

I'll merge this now.

Work is still needed to bring up a static mirror of lists.whatwg.org.

PR #270, opened 2019-12-18
Scrape lists.whatwg.org

foolip:


foolip:
@domenic r?

This might not be complete, but it's a good start from which to check if there's more that could be scraped that wget didn't catch.

foolip:
Merging this would deploy it to marquee, as in it would be there on disk, but it wouldn't be served anywhere. But it would allow me to experiment before the DNS change.

foolip:
I should also note that this makes the who repo much bigger. https://github.com/whatwg/misc-server/issues/107 is about fixing this, but if this repo would outlive moving static resources to a storage bucket, then it might not be a good idea to increase its size. If so, create a new repo just for this?

foolip:
Some rough projections before it's too late... this PR adds ~10k emails, i.e. files named `0*.html`, adding about 100MB. From https://github.com/whatwg/meta/issues/153#issuecomment-566980200 I'd guess we'd add at least 40k more, so maybe 400MB more.

I dunno, maybe this repo is fine, but I think I'll do this work in a separate repo first at least until we need to deploy it the first time.

@domenic or do you think even blowing up the repo 10x doesn't matter?

domenic:
I'm mostly just trying to compare to my existing experience with medium-sized repos like whatwg/html (with full history) and what we already have here. Anything similar to those seems OK.

foolip:
OK, I'll scrape web.archive.org and see how big the result is, then it's easier to decide.

foolip:
I'll work on this in https://github.com/foolip/whatwg.org/tree/scrape-lists, will delete admin.cgi.

annevk:
If we have all those email bodies, can we do some kind of automated search on W3C's infra to get the new permanent locations?

foolip:
I think we could, but the tricky part is determining what the original lists.whatwg.org URLs for those bodies should be, since those are the URLs we'd be trying to revive or redirect.

PR #284, opened 2020-02-17
Link to lists.whatwg.org from whatwg.org/mailing-list

foolip:
Everything that could be recovered from the Wayback Machine has been
restored, so there's no reason to link to it. The redirect mechanism
was never deployed.

### DIFF
--- a/whatwg.org/mailing-list
+++ b/whatwg.org/mailing-list
@@ -36,7 +36,10 @@ home.</p>
 
 <h3>Archives</h3>
 
-<p>A static archive is available at <a href="https://lists.whatwg.org/">lists.whatwg.org</a>.</p>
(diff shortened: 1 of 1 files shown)


foolip:
We could do that on the 404 page, which I'll send a separate PR for on top of https://github.com/whatwg/whatwg.org/pull/285. (Merging that will first delete it from the live site due to `--delete`.)

Do you think the 404 page is enough, or maybe do it here and point to it from the 404 page?

annevk:
Up to you, we can improve if there turns out to be a need.

foolip:
I think I'll put all the useful info on this page then, for one thing it can be linked to.

foolip:
Added "If you've found a broken link ..."

PR #285, opened 2020-02-17
Add a static copy of lists.whatwg.org

foolip:
This was created by scraping lists.whatwg.org while the DNS still
pointed at the original DreamHost setup, as well as extensive scraping
of web.archive.org, where everything that could be salvaged has been.
    
See the commit history and scripts for details of how this was done.


foolip:
Commit 691bf91d93f74b9872923d95c97952105ad7e4e1 merges in a standalone history from the lists.whatwg.org repo, similar to bcab19cde12921d76de3579a9580a0a47db83537. The best way to merge this would be either to push that commit to master, or to squash everything. The original history would be available via this PR.

This adds ~289M to the repo size, not as much as I originally thought.

annevk:
Let's squash it to keep linear history?

foolip:
Squashing is OK with me, will probably also reduce the size a little bit as some files changed over the course of the branch.

The useful way to review this is probably to read the commit messages in order and poke around on lists.whatwg.org to see what's still broken and needs fixing.

foolip:
@annevk can I go ahead and merge this? If you can't open https://github.com/whatwg/whatwg.org/pull/285/files I suppose there's also no way for you to approve this?

annevk:
Yeah I'm fine with you merging this. You suggested a set of things to do for review, but I haven't made the time and I'm not sure I will given that from a rudimentary check it appears okay. Is that okay?

foolip:
Sure, I think there's not that much risk here. I'll squash it all into a single commit, and then check to see that the resulting commit only had changes in lists.whatwg.org as expected.

foolip:
Alright, https://github.com/whatwg/whatwg.org/commit/2c69ab0349c4f13d02ec00d8bb0c326bbda2a6a6 added 103332 files, all in lists.whatwg.org/.

The deploy in https://travis-ci.org/whatwg/whatwg.org/builds/654327058 then failed because "The job exceeded the maximum log length, and has been terminated."

Ouch. I don't think restarting will help because all files are listed even if not changed, but I'll try it once to confirm.

foolip:
I can't find out how to restart the job in Travis, sent https://github.com/whatwg/whatwg.org/pull/286 to hopefully fix the problem though.

PR #286, opened 2020-02-24
Use `rsync --itemize-changes` to reduce log verbosity

foolip:
https://github.com/whatwg/whatwg.org/pull/285 added 103k files which
broke deploy on Travis due to exeeding the maximum log length.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -55,5 +55,7 @@ else
     eval "$(ssh-agent -s)"
     ssh-add deploy_key
     echo "$SERVER $SERVER_PUBLIC_KEY" > known_hosts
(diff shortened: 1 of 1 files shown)


foolip:
Nope, this [didn't work](https://travis-ci.org/whatwg/whatwg.org/builds/654338879). rsync is still copying everything, and going by [this guide](http://www.staroceans.org/e-book/understanding-the-output-of-rsync-itemize-changes.html) it's because the file modification time is different. That makes sense since it's a fresh checkout every time, but it's also pretty wasteful probably.

@annevk suggested in chat to redirect the log output to a file, but we'd need to upload that file as a build artifact for it to work.

Issue #287, opened 2020-02-24
Too many files for deploy to list them all


foolip:
https://github.com/whatwg/whatwg.org/pull/285 added 103k files which broke deploy on Travis due to exeeding the maximum log length. We have in total 108k files in the repo now.

https://github.com/whatwg/whatwg.org/pull/286 didn't work, and to get things working ASAP I'm going to suggest not logging the files at all.

Filing this issue to keep track of the problem.

annevk:
My suggestion is that we use --log-file=FILE and then scp that file (if it doesn't get synced) so it can be read over HTTP if someone is interested.

annevk:
See also #290.

foolip:
Thanks for finishing what I broke, @annevk!

PR #288, opened 2020-02-24
Avoid listing files entirely when deploying

foolip:
See https://github.com/whatwg/whatwg.org/issues/287.


foolip:
https://travis-ci.org/whatwg/whatwg.org/builds/654394605 passed so this fixed the problem, but the logs aren't very useful.

PR #289, opened 2020-02-24
Add back some rsync logging

### LINKED ISSUE #287: Too many files for deploy to list them all

annevk:
This also adds --stats for minimal logging in the CI log.

Closes #287 hopefully.

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -59,6 +59,6 @@ else
     # without exceeding log size limits:
     # https://github.com/whatwg/whatwg.org/issues/287
     rsync --archive --chmod="D755,F644" --compress --delete --stats --log-file="rsync-log.txt" --rsh="ssh -o UserKnownHostsFile=known_hosts" ./whatwg.org ./*.whatwg.org "deploy@$SERVER:/var/www/"
-    scp rsync-log.txt "deploy@$SERVER:/var/www/whatwg.org"
+    scp rsync-log.txt "deploy@$SERVER:/var/www/whatwg.org/"
     echo "Full rsync log at https://whatwg.org/rsync-log.txt"
 fi
--- a/deploy.sh
+++ b/deploy.sh
@@ -58,5 +58,7 @@ else
     # --verbose isn't used because there are too many files to list them all
     # without exceeding log size limits:
     # https://github.com/whatwg/whatwg.org/issues/287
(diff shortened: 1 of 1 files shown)


deploy.sh
```diff
@@ -58,5 +58,7 @@ else
     # --verbose isn't used because there are too many files to list them all
     # without exceeding log size limits:
     # https://github.com/whatwg/whatwg.org/issues/287
-    rsync --archive --chmod="D755,F644" --compress --delete --rsh="ssh -o UserKnownHostsFile=known_hosts" ./whatwg.org ./*.whatwg.org "deploy@$SERVER:/var/www/"
+    rsync --archive --chmod="D755,F644" --compress --delete --stats --log-file="rsync-log.txt" --rsh="ssh -o UserKnownHostsFile=known_hosts" ./whatwg.org ./*.whatwg.org "deploy@$SERVER:/var/www/"
+    scp rsync-log.txt "deploy@$SERVER:/var/www/whatwg.org"
```
annevk:
```suggestion
    scp rsync-log.txt "deploy@$SERVER:/var/www/whatwg.org/"
```

foolip: Does it work without this suggestion, I guess not?

deploy.sh
```diff
@@ -58,5 +58,7 @@ else
...
+    echo "Full rsync log at https://whatwg.org/rsync-log.txt"
```
foolip: Oh, well that works :)

annevk:
I think it works without because whatwg.org exists, but if it didn't it would rename I think rather than throw an error.

annevk:
One worry I have is that the log file on the server will first be send back due to rsync before being overwritten. But that's probably okay?

foolip:
The remote rsync-log.txt should be deleted by the first command because there won't be a whatwg.org/rsync-log.txt file. That seems harmless.

annevk:
Oh my. The scp doesn't know about the known hosts.

PR #290, opened 2020-02-25
Ensure scp knows the target server as well

annevk:
I'm pretty sure this should work.


foolip:
Yep, it worked: https://travis-ci.org/whatwg/whatwg.org/builds/655103430

annevk:
Cool, and https://whatwg.org/rsync-log.txt also works.

PR #292, opened 2020-03-06
Make SG footers consistent

annevk:
Closes https://github.com/whatwg/sg/issues/78.

Not a great solution, but curious to see what people have to suggest.

### DIFF
--- a/convert-policy.py
+++ b/convert-policy.py
@@ -1,4 +1,5 @@
 #!/usr/bin/env python
+# -*- coding: UTF-8 -*-
 
 import codecs
 import markdown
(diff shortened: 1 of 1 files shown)


annevk:
Unfortunately this is wrong as Code of Conduct has a more complicated footer. It does seem like this would work with `commonmark`. I suspect we'd want to keep the `<hr>` in the Markdown source though as `<small>` is not picked up. But then we'd only have to replace `<hr><footer>` with `<footer>` here so that's worth pursuing I think.

annevk:
I suspect the fix here will be as shown in #296. Change all the Markdown files once to include `<footer>` and replace `<hr>` followed by `<footer>` with `<footer>`.

PR #293, opened 2020-03-06
Slight readability improvements

travisleithead:
There are a few things that really make review drafts hard to read for me:
* The browser's default serif font (no font-family set for the body)
* Lack of page padding and bordered box padding 

I'm hopeful that the following small tweaks will be acceptable to improve basic readability within the existing purpose and intent of the review draft format.

### DIFF
--- a/resources.whatwg.org/review-draft.css
+++ b/resources.whatwg.org/review-draft.css
@@ -1,5 +1,5 @@
 pre.idl, pre.asn, pre.css, dl.domintro, .note, .warning, .example {
- border: 1px solid black;
+ border: 1px solid black; padding: 0.5em;
 }
 
 pre.idl::before, pre.asn::before, pre.css::before, dl.domintro::before, .note::before, .warning::before, .example::before {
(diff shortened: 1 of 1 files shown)


domenic:
I'd rather not make these changes. Review drafts are intentionally unformatted. To the extent your browser allows changing its default fonts/margins/etc., either through settings or dev tools, you can. But by default review drafts are meant to be a record for patent review, and not more.

travisleithead:
May not change your mind, but worth considering that with the MoU in place, at least DOM and HTML review drafts will be more than just a record for patent review, they will be the official "stable" editions that W3C links will redirect to (likely increasing readership dramatically and potential similar feedback).

domenic:
Sure, but hopefully folks arriving there will quickly see the yellow warning box and figure out what the best way to read the HTML/DOM Standards are.

othermaciej:
I don't think we need to go out of our way to make the documents hard to read, at least not as much as currently. The warning box is enough.

There is one purposes of the documents we all endorse, recording a specific version for which there are commitments to license essential patent claims royalty-free. Both under WHATWG IPR Policy and now under W3C Patent Policy. For that purpose, it's good to be able to read the content easily and understand intent.

I wish we could find a less obtrusive way to discourage using these as a reference. Maybe a discouraging watermark background.

annevk:
If we do this, shouldn't we move the relevant styles from standard.css to spec.css and remove any duplicative rules from review-draft.css?

Issue #291, opened 2020-03-06
convert-policy.py and Markdown


annevk:
* The only affect of `from mdx_partial_gfm import PartialGithubFlavoredMarkdownExtension` appears to be the auto-linking of the email address in the Code of Conduct. We could consider replacing it with something simpler that handles email addresses.
* Adopting `commonmark` would require some restructuring of the code as it doesn't like some of the pre-processing we do. It ends up wrapping our headings with paragraphs and also encodes a bunch of " as `&quot;`.

It's not clear that making either is worth it.

domenic:
I think Commonmark is worth a lot just in being the same as GitHub and helping the Markdown ecosystem converge toward a single spec.

annevk:
Could it be that the Python `commonmark` package is wrong? https://spec.commonmark.org/ behaves very similar to the `markdown` package we are using, especially for input such as
```
<h3>...</h3>
test
```
or
```
<h3>...</h3>

test
```
whereas if I try that locally with the `commonmark` package the `<h3>` ends up wrapped in a `<p>`.

domenic:
That could be possible, if distressing :(. Although as I noted in https://github.com/whatwg/sg/pull/92#issuecomment-596056668 I'm unsure why our input contains `<h3>`s anyway, instead of `###`.

annevk:
That's what convert-policy does. We could possibly change it to do post-HTML-conversion changes. That would also fix #145. But I don't think I've really seen where commonmark and markdown give different results yet. Other than that commonmark didn't do email address auto-linking. (Perhaps it needs to be wrapped inside angle brackets.)

Issue #264, opened 2019-09-23
convert-policy.py and Python 3


annevk:
I'm guessing we need to update this or will Travis et al support Python 2.7 for the foreseeable future?

domenic:
Various things have gotten more aggressive about removing Python 3 support. Also, [our builds have started failing with weird encoding errors](https://travis-ci.org/whatwg/whatwg.org/builds/659206562#L206), even on Python 2. So, we should probably do something here sooner rather than later.

Issue #145, opened 2017-12-21
convert-policy.py loses Markdown links in headers


othermaciej:
convert-policy.py loses Markdown links in headers. This doesn't currently affect any WHATWG policy documents, but it would be nice to fix at some point.

annevk:
#295 pretty much fixes this, except that the generated IDs are bad. `header_text_to_id` will also need to strip markup. Otherwise you get
```html
<h3 id="conduct-a-hrefhttpsexmamplecomtesta">Conduct <a href="https://exmample.com/">test</a><a class="self-link" href="#conduct-a-hrefhttpsexmamplecomtesta"></a></h3>
```
for
```markdown
## Conduct [test](https://exmample.com/)
```

PR #294, opened 2020-03-07
Testing the future of markdown

annevk:
This will mainly help with reviewing some planned changes to convert-policy.py to address #291.

PR #295, opened 2020-03-07
Use CommonMark in convert-policy.py

annevk:
This fixes a few subtle issues and is a prerequisite for fixing more.

### DIFF
--- a/convert-policy.py
+++ b/convert-policy.py
@@ -1,7 +1,5 @@
 #!/usr/bin/env python
-# -*- coding: UTF-8 -*-
 
-import codecs
 import commonmark
 import re
 
@@ -26,6 +24,7 @@ def ascii_lower(str):
 
 
 def header_text_to_id(header_text):
+    header_text = header_text.replace("&quot;", "")
     punctuation_regexp = r'[^\w\- ]'
     header_id = ascii_lower(header_text)
     header_id = re.sub(punctuation_regexp, '', header_id)
@@ -63,8 +62,7 @@ def preprocess_markdown(policy_markdown, mapping_pairs):
 
 
 def postprocess_html(policy_html, template, title):
-    result = policy_html.replace("&quot;", "\"")
-    result = template.replace("@POLICY_GOES_HERE@", result)
+    result = template.replace("@POLICY_GOES_HERE@", policy_html)
     result = result.replace("@TITLE_GOES_HERE@", title)
     result = add_header_anchors(result)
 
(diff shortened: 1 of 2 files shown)


annevk:
https://github.com/whatwg/sg/pull/104 should probably land first. The commits in #294 can be used to review the effect of this. I ended up cleaning up the code a bit further compared to that, but I promise there's no difference in output.

annevk:
I'm not sure if this fixes #145 as regular expressions confuse me a ton.

It does seem that switching to using Python 3 is fine and this would fix #264 therefore.

annevk:
This should close #291 as well.

convert-policy.py
```diff
@@ -1,8 +1,8 @@
 #!/usr/bin/env python
+# -*- coding: UTF-8 -*-
```
domenic: Is this necessary still? It seemed surprising to me that it was as we don't seem to use any Unicode literals in the source text...

convert-policy.py
```diff
@@ -55,21 +54,19 @@ def rewrite_defs(policy_markdown):
...
-    result = add_header_anchors(result)
     result = rewrite_defs(result)
-    result = fix_nested_lists(result)
-    result = avoid_link_false_positives(result)
+
+    return result
+
+
+def postprocess_html(policy_html, template, title):
+    result = policy_html.replace("&quot;", "\"")
```
domenic: This seems dangerous as a blanket transformation. Should we just let Commonmark do its thing?

convert-policy.py
```diff
@@ -97,11 +94,10 @@ def main():
...
 
-        policy_html = markdown.markdown(preprocessed_policy_markdown, extensions=[PartialGithubFlavoredMarkdownExtension()])
+        policy_html = commonmark.commonmark(preprocessed_policy_markdown)
 
-        final_policy_html = template.replace("@POLICY_GOES_HERE@", policy_html)
-        final_policy_html = final_policy_html.replace("@TITLE_GOES_HERE@", title)
+        postprocessed_policy_html = postprocess_html(policy_html, template, title)
 
-        codecs.open("whatwg.org/" + link, "w", encoding="utf-8").write(final_policy_html)
+        codecs.open("whatwg.org/" + link, "w", encoding="utf-8").write(postprocessed_policy_html)
```
domenic: I wonder why we're using this instead of just `open()`.

It looks like this might be a Python 2ism so we can just move to open (and remove the codecs import): https://stackoverflow.com/questions/5250744/difference-between-open-and-codecs-open-in-python

convert-policy.py
annevk: It's not needed, but seems better to declare?

convert-policy.py
annevk: No, that will create really messy IDs. This is needed for a minimal diff with what we have today. I guess the main future problem would be when setting the title attribute through Markdown?

convert-policy.py
annevk: Yeah I can do that.

convert-policy.py
domenic: Hmm, I don't know the best practices, but it feels like a workaround.

https://stackoverflow.com/questions/14083111/should-i-use-encoding-declaration-in-python-3 says it's not needed because the default is UTF-8, at least in Python 3.

convert-policy.py
domenic: Yeah, or any raw HTML uses.

I'm OK with a non-minimal diff, but if this will affect ID generation that's not great. Maybe the ID generation code should do this transformation itself before generating the ID?

annevk: Could you review https://github.com/whatwg/whatwg.org/pull/294/commits/f51ca893eac8620347a2b0f7147eb36cf1f59e11 to ensure that's all that's changed? I guess another thing to consider is using better quotation marks in the source material, but that might well require changes to header to ID conversion as well.

domenic: Reviewed! Looks good to me too; thanks for going the extra step.

annevk:
Feedback addressed.

PR #298, opened 2020-03-07
Editorial: remove dead code from convert-policy.py

annevk:
Between 4d34512aeb618c8753e317b220a7937c2161fb11 and https://github.com/whatwg/sg/pull/77 this became obsolete I think.


annevk:
Turns out Python has asserts!

PR #300, opened 2020-03-10
Specify the Python version to use on Travis (3.8)

foolip:
The default is now 3.6. The important thing here is to make it clear
that Python 3 is used in CI, since just `pip` and `python` are used
in scripts, which usually point to Python 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
 language: python
+python: "3.8"
 
 env:
   global:


annevk:
I kinda like that we don't have a number here to bump every now and then. So I'll let @domenic decide whether to take this.

annevk:
There's one potential advantage in that only Python 3.7 and above have guaranteed ordering for dictionaries. I believe Bikeshed also requires 3.8, though we don't use it directly.

PR #301, opened 2020-03-10
Make it clear that lists.whatwg.org is a static archive

foolip:
Goes with https://github.com/whatwg/misc-server/pull/126.


foolip:
I propose to let /listinfo.cgi simply be 404 (like /admin.cgi) since there isn't anything to preserve. The alternative would be to redirect to /.

annevk:
FWIW, I argued in https://github.com/whatwg/misc-server/pull/126 for not letting these 404 as they are the pages you end up in if you navigate to lists.whatwg.org today so folks might well have bookmarks or links to them. If nobody finds that compelling I'm happy to let it be though.

domenic:
Redirecting the main `/listinfo.cgi/` (maybe `/listinfo.cgi` too?) to some other landing page makes sense. The sub-pages make more sense as a 404. (But I don't feel strongly in any direction.)

foolip:
Alright, I'll add some redirects back in https://github.com/whatwg/misc-server/pull/126.

foolip:
This has been deployed now. https://whatwg.org/mailing-list gets me this error in my usual Chrome:
> This site can’t be reached
> The webpage at https://whatwg.org/mailing-list might be temporarily down or it may have moved permanently to a new web address.

It works in incognito though. It's probably because 301 is cached somewhere. Not trying to fix it because the problem will go away with time :)

annevk:
Is that due to the service worker?

foolip:
I didn't suspect that, but looks like it is!
```
website-service-worker.js:47 Could not refresh the cache for https://whatwg.org/mailing-list Error: https://whatwg.org/mailing-list is responding with 0
    at refreshCacheFromNetworkResponse (website-service-worker.js:58)
    at website-service-worker.js:45
```

foolip:
The bug is probably in https://github.com/whatwg/whatwg.org/blob/master/resources.whatwg.org/website-service-worker.js then.

PR #303, opened 2020-03-14
Use SG's db.json as source of truth

annevk:
See whatwg/sg#111 for context.

Changes to the status quo:

* biblio.json: LOADER no longer has `"status": "Living Idea"`. Two author names are now sorted in lexicographical order.
* spec.whatwg.org / idea.whatwg.org: A lot of whitespace changes, which we could limit a bit more, but I like the new style better...


annevk:
(Note that this fails Travis CI because db.json doesn't exist yet.)

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+
+def standard_or_idea_to_biblio(document):
+    return (document["reference"], {
+        "authors": [author_to_biblio(author) for author in document["authors"]],
+        "href": document["href"],
+        "title": document["name"]
+    })
+
+def author_to_biblio(person):
+    return person["name"]
```
domenic: Nit: move after `biblio_item_to_biblio` to keep `biblio_item_to_biblio` and `standard_or_idea_to_biblio` together?

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+
+    output = template.replace("@TITLE_GOES_HERE@", "Standards")
+    output = output.replace("@CONTENT_GOES_HERE@", content)
+    output = output.replace("<a href=\"https://spec.whatwg.org/\">Standards</a>", "<a>Standards</a>")
+
+    open("spec.whatwg.org/index.html", "w").write(output)
+
+def standard_or_idea_to_html(document):
+    output = "\n <dt><a href=\"{0}\">{1}</a></dt>".format(document["href"], document["name"])
+    return output + "\n <dd>{}</dd>\n".format(commonmark.commonmark(document["description"])[:-1])
```
domenic: Nit: this might be nicer:

```python
# TODO add a comment for the [:-1] here
description = commonmark.commonmark(document["description"])[:-1]

return "\n <dt><a href=\"{0}\">{1}</a></dt>\n <dd>{2}</dd>\n"
       .format(document["href"], document["name"], description)
```

Either way, the `[:-1]` is confusing and a comment explaining it would probably help.

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+    content += "</dl>\n\n"
+    content += copyright
+
+    output = template.replace("@TITLE_GOES_HERE@", "Ideas")
+    output = output.replace("@CONTENT_GOES_HERE@", content)
+
+    open("idea.whatwg.org/index.html", "w").write(output)
+
+
+def create_workstreams(db, template, copyright):
```
domenic: Oof, templating in Python shows its stress here.

I'm happy to keep this approach since it's a big improvement over what we have now. But if you are interested, using https://github.com/wbond/pybars3 might make this nicer (especially for this page, but also for the others). It could replace all the templating with separate `.hbs` template files.

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+    content = """<p>This document contains an official <a href="/sg-agreement#steering-group">Steering Group</a>-maintained accounting of the WHATWG workstreams, their responsible editors, and their associated standards.</p>
+
+<p>Note that per the <a href="/workstream-policy">Workstream Policy</a>, the official <a href="/workstream-policy#editor">Editors</a> listed here may have delegated responsibilities and editing duties to deputy editors for any given <a href="/workstream-policy#living-standard">Living Standard</a>. As such, the <a href="/workstream-policy#editor">Editor</a> position listed here is more about "who is the official liaison with the <a href="/sg-agreement#steering-group">Steering Group</a>" and less about "who is editing the document". To get an accounting of who is primarily responsible for day-to-day specification work, you'll be better served by checking the Acknowledgments section of the <a href="/workstream-policy#living-standard">Living Standard</a> in question.</p>"""
+    content += "\n\n<dl>"
```
domenic: Since you already have a multiline string I'd just insert the `\n\n<dl>` into it?

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+    for workstream in db["workstreams"]:
+        content += "\n <dt>{}</dt>".format(workstream["name"])
+        content += "\n <dd>"
+        content += "\n  <p><strong>Scope:</strong> {}</p>".format(commonmark.commonmark(workstream["scope"])[3:-5]) # Strip leading <p> and trailing </p>\n
```
domenic: This list of paragraphs with `<strong>` items feels like a dl in disguise.

convert-sg-db.py
```diff
@@ -0,0 +1,131 @@
...
+    output = output.replace("@CONTENT_GOES_HERE@", content)
+
+    open("whatwg.org/workstreams", "w").write(output)
+
+
+def main():
+    template = open("site-template.html", "r", encoding="utf-8").read()
+    copyright = """<footer>
+ <p><small>Copyright © 2018 WHATWG (Apple, Google, Mozilla, Microsoft). This work is licensed under a <a rel="license" href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0 International License</a>.</small></p>
+</footer>"""
```
domenic: Can we put the copyright in the template?

convert-sg-db.py
annevk: I suppose we could do that. Do we still want to differentiate as much between singular and plural then? Or use multiple `<dd>`s when there happen to be multiple?

convert-sg-db.py
annevk: Then we'd have to strip it in convert-policy.py.

annevk: I suppose we could run a replace on _template_ locally though instead of passing it everywhere.

convert-sg-db.py
domenic: Ooh, interesting. I think just grammatically I like keeping `Editor:` vs. `Editors:` but yeah, we could simplify the markup structure in the way you suggest.

convert-sg-db.py
domenic: Hmm yeah.

Also 2018 seems wrong. We can probably just use current year for these?

annevk: I suppose we could, at least until that issue about what year to use is resolved.

convert-sg-db.py
annevk: I've kept this for now as less dependencies is also nice and I'm not very familiar with templating packages. Can always change this if we make things more complicated.

convert-sg-db.py
```diff
@@ -0,0 +1,121 @@
...
+        for standard in workstream["standards"]:
+            content += standard_or_idea_to_html(standard)
+    content += "</dl>\n\n<p>The WHATWG also works on a <a href=\"https://idea.whatwg.org/\">number of ideas</a> that aspire to become standards one day.</p>\n\n"
+
+    template = template.replace("<a href=\"https://spec.whatwg.org/\">Standards</a>", "<a>Standards</a>")
+    write_site_template("spec.whatwg.org/index.html", template, "Standards", content)
+
+def standard_or_idea_to_html(document):
+    output = "\n <dt><a href=\"{0}\">{1}</a></dt>".format(document["href"], document["name"])
+    return output + "\n <dd>{}</dd>\n".format(commonmark.commonmark(document["description"])[:-1]) # Strip trailing \n
```
domenic: I still think splitting this into two lines is a bit weird, and an intermediate variable would clean things up.

convert-sg-db.py
```diff
@@ -0,0 +1,121 @@
...
+
+def write_site_template(filename, template, title, content):
```
domenic: Move this down above main to keep the `create_`s together?

convert-sg-db.py
```diff
@@ -0,0 +1,121 @@
...
+
+
+def main():
+    template = open("site-template.html", "r", encoding="utf-8").read()
+    copyright = """<footer>
+ <p><small>Copyright © {} WHATWG (Apple, Google, Mozilla, Microsoft). This work is licensed under a <a rel="license" href="https://creativecommons.org/licenses/by/4.0/">Creative Commons Attribution 4.0 International License</a>.</small></p>
+</footer>""".format(str(date.today().year))
+    template = template.replace("@CONTENT_GOES_HERE@", "@CONTENT_GOES_HERE@" + copyright)
+    db = json.load(open("../sg/db.json", "r"))
+    create_biblio(db)
```
domenic: Newline before here to separate things nicely?

convert-sg-db.py
```diff
@@ -0,0 +1,121 @@
...
+        for editor in workstream["editors"]:
+            content += """\n   <dd>{0}, <a href="mailto:{1}">{1}</a></dd>""".format(editor["name"], editor["email"])
+        content += "\n   <dt>Standard{}</dt>".format("" if len(workstream["standards"]) == 1 else "s")
+        for standard in workstream["standards"]:
+            content += """\n   <dd><a href="{0}">{1}</a></dd>""".format(standard["href"], standard["name"])
+        content += "\n  </dl>"
+        content += "\n </dd>\n"
+    content += "</dl>\n\n"
+
+    write_site_template("whatwg.org/workstreams", template, "Workstreams", content)
```
domenic: I think it would be nicer if all the I/O was in main() and the create_ functions were just pure functions that returned a string.

convert-sg-db.py
```diff
@@ -0,0 +1,121 @@
...
+    content += "<dl>"
+    for workstream in db["workstreams"]:
+        for standard in workstream["standards"]:
+            content += standard_or_idea_to_html(standard)
+    content += "</dl>\n\n<p>The WHATWG also works on a <a href=\"https://idea.whatwg.org/\">number of ideas</a> that aspire to become standards one day.</p>\n\n"
+
+    template = template.replace("<a href=\"https://spec.whatwg.org/\">Standards</a>", "<a>Standards</a>")
+    write_site_template("spec.whatwg.org/index.html", template, "Standards", content)
+
+def standard_or_idea_to_html(document):
```
domenic: Move this down to be below create_idea and create_spec, instead of in between them?

convert-sg-db.py
annevk: I put them in the order you encounter them in (except for main, which is at the bottom).

convert-sg-db.py
annevk: I think it's okay. It's also consistent with the other commonmark caller.

convert-sg-db.py
annevk: Same as above, it's mostly order of appearance.

convert-sg-db.py
domenic: I think that's significantly less helpful than logical ordering, and would really prefer otherwise.

annevk: What is logical ordering?

domenic: Grouping related things, especially things of the same inputs/return values, together. Instead of interspersing helpers between them.

annevk: Is this documented somewhere? I looked at PEP8 and it only seems to care about import order.

domenic: No, this isn't Python specific, it's just a general programming style thing. A quick Googling shows things like https://google.github.io/styleguide/cppguide.html#Declaration_Order or https://kotlinlang.org/docs/reference/coding-conventions.html#class-layout

convert-sg-db.py
```diff
@@ -0,0 +1,122 @@
...
+    content += "</dl>\n\n<p>The WHATWG also works on a <a href=\"https://idea.whatwg.org/\">number of ideas</a> that aspire to become standards one day.</p>\n\n"
+
+    template = template.replace("<a href=\"https://spec.whatwg.org/\">Standards</a>", "<a>Standards</a>")
+    return wrap_in_site_template(template, "Standards", content)
+
+def standard_or_idea_to_html(document):
+    output = "\n <dt><a href=\"{0}\">{1}</a></dt>".format(document["href"], document["name"])
+    return output + "\n <dd>{}</dd>\n".format(commonmark.commonmark(document["description"])[:-1]) # Strip trailing \n
+
+def wrap_in_site_template(template, title, content):
```
foolip, line 66: https://github.com/whatwg/whatwg.org/commit/4bcc3c3e4b5c5abc176ef447f96e7f12c5cb615c reveals a problem in idea.whatwg.org/index.html, where the link will be broken because site-template.html assumes the result will be on whatwg.org and links /.

annevk:
Thanks, do I need to address those nits before the whatwg/sg part can be reviewed or can I wait for that to happen before fixing all the things?

deploy.sh
```diff
@@ -8,6 +8,7 @@ header() {
 header "Importing content from whatwg/sg and making it suitable for whatwg.org"
 git clone --depth=1 https://github.com/whatwg/sg sg
 ./convert-policy.py
+./convert-sg-db.py
```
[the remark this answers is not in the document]
annevk: I ended up doing the simple thing, but I could see allowing it to be optionally passed in. Let me know if it's important.

convert-sg-db.py
annevk: I addressed this by always using the full link. This is a bit redundant for some URLs, but does work kind of nice if you end up hosting the same content on another URL for preview reasons. 😊

foolip: Yep, that works!

deploy.sh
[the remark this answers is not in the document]
foolip: Nah, not important enough.

annevk:
FWIW, if you want to do a reordering commit that makes it more logical to you that seems fine to me.

domenic:
I pushed a reordering commit, a Windows fix (ugh), and a style fix for https://github.com/whatwg/sg/pull/111#issuecomment-617935515. We could also switch to `<p>` and `<strong>` instead of keeping the `<dl>` structure.

domenic:
Also, I guess we should probably make the build pass before merging this, right? I think it's missing a "clone the sg repo" step...

annevk:
The reason the build script fails is because the whatwg/sg PR has not landed and therefore it does not have the necessary files (`db.json` in particular, but now also `Workstreams Introduction.md`).

And yeah, once we merge the whatwg/sg PR we should rerun this to see if it goes green.

whatwg.org/style/subpages.css
```diff
@@ -74,6 +74,12 @@ dl > dt { margin: 1em 0 0 0; }
 dl > dd { margin: 0.25em 0 0.25em 2em; }
 ol.dated { list-style: disc; }
 
+dl.compact > div { margin: 1em 0; }
+dl.compact > div > dt { display: inline; margin: 0; font: inherit; font-weight: bold; }
+dl.compact > div > dt::after { content: ": "; }
+dl.compact > div > dd { display: inline; margin: 0; }
+dl.compact > div > dd:not(:last-of-type)::after { content: "; "; }
```
annevk, line 81: Does this mean we need to bump `cacheKeyToUse` in https://resources.whatwg.org/website-service-worker.js?

PR #309, opened 2020-03-18
Use *-snapshot.svg logos for Commit Snapshots

annevk:
Closes https://github.com/whatwg/meta/issues/88.


annevk:
A variant of this worked locally, but haven't tested this yet. Reasonable approach or should we pass this information into Bikeshed somehow?

domenic:
It's not terrible, but I think it'd be better to do this similar to how we do RDs, which is https://github.com/tabatkins/bikeshed/blob/master/bikeshed/boilerplate/whatwg/computed-metadata-RD.include . It might be as simple as adding `computed-metadata-LS-COMMIT.include`.

(Side note: I think we could clean up a good amount of the snapshot infrastructure using similar computed metadata techniques that make it similar to RDs. https://github.com/whatwg/whatwg.org/blob/75b4d979029558e4172f51bac43eeecf35b13e04/resources.whatwg.org/build/deploy.sh#L106-L110 could probably be replaced with `-F md-status=LS-COMMIT -F md-Text-Macro="SHA $SHA"`.)

PR #311, opened 2020-03-30
Change MDN-annotation styles to Bikeshed styles

sideshowbarker:
This changes add CSS styles for the MDN annotations generated by Bikeshed. The change also removes the existing CSS styles we were using for the MDN annotations generated by Wattsi for the HTML standard.

(A separate change to the Wattsi sources will revise the element markup and attribute markup generated by Wattsi such that it matches what Bikeshed generates — allowing the CSS styles in this change to be be used/shared among the HTML standard and all other WHATWG standards.)

### DIFF
--- a/resources.whatwg.org/standard-shared-with-dev.css
+++ b/resources.whatwg.org/standard-shared-with-dev.css
@@ -93,7 +93,227 @@ c-[vg] { color: #0077aa } /* Name.Variable.Global */
 c-[vi] { color: #0077aa } /* Name.Variable.Instance */
 c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
 
-/* MDN margin annotations */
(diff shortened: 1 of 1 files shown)


domenic:
We *may* want to support both variants, if feasible, for a couple months. E.g. I'm worrying about if people have the CSS cached but get the HTML fresh. But I'm OK just going for it, if we think that's easier.

sideshowbarker:
> We _may_ want to support both variants, if feasible, for a couple months. E.g. I'm worrying about if people have the CSS cached but get the HTML fresh. But I'm OK just going for it, if we think that's easier.

I think their might be no harm in keeping the existing ones after we merge the new ones. I *think* there are no conflicts, but I’ll doublecheck to make sure. We’d just need to remember (somehow) that N weeks from now we should go back in and remove the old ones.

PR #316, opened 2020-05-06
Do not italicize <pre> inside notes

domenic:
Inspired by https://github.com/whatwg/url/pull/495.

Before:

image

After:

image


annevk:
This is probably the better PR, though .note in front isn't needed in this case (might be slightly faster).

domenic:
Hmm, I think I like yours more, in case there are other italic cases (or we add any in the future).

PR #317, opened 2020-05-06
standard.css: pre { font-style: normal }

annevk:
This avoids a .note italicizing it: https://github.com/whatwg/url/pull/495.


annevk:
Probably not a big enough change to invalidate service workers over?

PR #318, opened 2020-05-12
standard.css: code { font-style:normal }

annevk:
This would make the warning in https://encoding.spec.whatwg.org/#dom-textdecoder-decode look a bit better (to me).


resources.whatwg.org/standard.css
```diff
@@ -48,7 +48,7 @@ body { margin: 0 auto; padding: 0 2.5em 2em 2.5em; max-width: 80em; background:
   .selected-text-file-an-issue { left: 0; right: auto; text-align: left; }
 }
 
-code { color: #666666; }
+code { color: #666666; font-style:normal; }
```
domenic, line 51:
```suggestion
code { color: #666666; font-style: normal; }
```

PR #321, opened 2020-06-10
Deploy to marquee using GitHub actions

foolip:


annevk:
To be clear, I'm convinced it's worth going in this direction. Feel free to land this.

foolip:
I've filed https://github.com/whatwg/meta/issues/173 to track the migration for all of our repositories. Before I land this I'll prepare a PR for a few other repos too, to shake out any unexpected issues with the different types of setups we have.

foolip:
Alright, I'll make a backup of marquee and deploy this now, to see that this doesn't blow up the world before proceeding with other things that deploy to marquee.

foolip:
https://github.com/whatwg/whatwg.org/runs/762311824 worked after I fixed the secrets.

annevk:
I think this regressed us a tiny bit as before we had shellcheck run always and deploy only run on master.

foolip:
You're right, I'll fix that.

foolip:
Fixed in https://github.com/whatwg/whatwg.org/pull/323.

PR #322, opened 2020-06-11
Enable specs to deploy using GitHub Actions

foolip:
Part of https://github.com/whatwg/meta/issues/173.

### DIFF
--- a/resources.whatwg.org/build/README.md
+++ b/resources.whatwg.org/build/README.md
@@ -12,18 +12,45 @@ The `deploy.sh` script is used by most WHATWG standards and is meant to run eith
 - Running the [Nu HTML checker](http://checker.html5.org/) on the build results
 - Deploying the build results to the WHATWG web server
 
-For non-local deploys, it is dependent on the following environment setup:
+For Travis CI deploys, it is dependent on the following environment setup:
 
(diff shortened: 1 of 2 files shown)


resources.whatwg.org/build/README.md
```diff
@@ -12,19 +12,46 @@ The `deploy.sh` script is used by most WHATWG standards and is meant to run eith
...
+
+```yaml
+name: deploy
+on:
+  push:
+    branches:
+      - master
+jobs:
+  deploy:
+    runs-on: ubuntu-20.04
```
domenic, line 41: We've been using Debian elsewhere; any particular reason to go with Ubuntu here?

foolip: https://help.github.com/en/actions/reference/virtual-environments-for-github-hosted-runners#supported-runners-and-hardware-resources is the set of GitHub-provided runners. We could provide our own Docker image and lock everything down, but I actually think it's not necessary for the relatively plain dependencies that we have.

foolip:
I fixed the shellcheck failure (missing quotes) and rebased. When I've merged this and it's deployed to resources.whatwg.org, I'll merge https://github.com/whatwg/fullscreen/pull/174 to see if I broke the existing Travis builds.

PR #323, opened 2020-06-11
Run deploy.sh on PRs as well

foolip:
Also take this chance to simplify the TODO waiting for Bash 4.4. Ubuntu
20.04 appears to have Bash 5.

This regressed in https://github.com/whatwg/whatwg.org/pull/321.

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -1,8 +1,11 @@
 name: deploy
 on:
+  pull_request:
+    branches:
+    - master
   push:
     branches:
-      - master
+    - master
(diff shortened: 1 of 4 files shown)


domenic:
> TODO: if Travis CI ever gets Bash 4.4, we can use the simpler version at https://stackoverflow.com/a/54561526/3191.

might be worth investigating.

foolip:
I've fixed the Bash 4.4 TODO and confirmed that it's still finding and validating files from both groups:
https://github.com/whatwg/whatwg.org/pull/323/checks#step:9:41

PR #325, opened 2020-06-12
Tolerate no trailing .git when determining SHORTNAME

foolip:


foolip:
Yes, this is an hilarious amount of escaping!

annevk:
Why is this needed?

I'm wondering if it would be better to hardcode the couple of exceptions we have here with respect to shortname in factory.json and propagate it as input parameter.

annevk:
Sorry, I was thinking about the wrong thing. We should probably indicate we can clean this up post-transition.

foolip:
I'll send a follow-up that does something much simpler for GitHub Actions. The other path will then be for local, where it actually makes sense to handle both cases, and will continue to work on Travis.

foolip:
Sent https://github.com/whatwg/whatwg.org/pull/326.

PR #326, opened 2020-06-12
Make SHORTNAME detection on GitHub Actions much simpler

foolip:
The complicated thing is still used locally and on Travis.

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -6,10 +6,6 @@ set -o nounset
 # https://github.com/whatwg/whatwg.org/tree/master/resources.whatwg.org/build.
 # See README.md for documentation.
 
-SHORTNAME=$(git config --local remote.origin.url | sed -n 's#.*/\([^.]*\)\(\.git\)\?#\1#p')
-INPUT_FILE=$(find . -maxdepth 1 -name "*.bs" -print -quit)
-
-WEB_ROOT="$SHORTNAME.spec.whatwg.org"
 COMMITS_DIR="commit-snapshots"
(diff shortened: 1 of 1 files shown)


foolip:
Here I also organized things a bit into constants, environment variables and local variables.

annevk:
So we basically have this because of compat/console/streams not using their shortname as their .bs name, right? @domenic any thoughts on renaming them instead?

domenic:
Renaming is quite disruptive and loses the entire file history in the GitHub UI. I think adding a single line in the build script is a very small price to pay.

(I'm not really sure why we'd expand that single line into 5 lines as this PR does, though...)

annevk:
Well, you wouldn't lose history for the links we advocate, e.g., https://github.com/whatwg/compat/commits, and I think blame isn't impacted either, but that's fair.

foolip:
> I'm not really sure why we'd expand that single line into 5 lines as this PR does, though...

Things work without this change, but interpreting that `sed` command is very hard. This is a follow-up to https://github.com/whatwg/whatwg.org/pull/325 to make it more obvious why it's like that (in the future: for local builds only). But if this just looks more convoluted, I could either:
- abandon this
- just use the directory name in the else branch

annevk:
I would abandon, but explain the complexity in a comment.

domenic:
Given that the sed command still exists, it doesn't feel like a win to just add a second path where the sed command is bypassed. Now one has to understand `complicated thing + simple thing`, which seems slightly worse than having to understand `complicated thing` by itself.

foolip:
I've sent https://github.com/whatwg/whatwg.org/pull/327 instead.

PR #330, opened 2020-06-14
Use `android-webview` instead of the `android-webview-beta` package

alrra:
The `Android WebView` was (is?) actually part of the Android Beta program, but now there is `Android WebView Beta`, so the `browser-logos` packages have been updated accordingly.

Ref: 

* https://play.google.com/store/apps/details?id=com.google.android.webview
* https://play.google.com/store/apps/details?id=com.google.android.webview.beta


annevk:
Thanks!

PR #332, opened 2020-06-16
Skip deploy step for forks of our spec repos

foolip:
https://github.com/shivanigithub/fetch/runs/776676042 failed of
attempting to deploy from a fork where the secrets aren't present.


foolip:
This was reported by @annevk. r?

annevk:
I'm not familiar with the whatwg/* syntax to match a string.

foolip:
Me neither, I found it at https://stackoverflow.com/a/2172367 and tested it locally, seems to work.

PR #333, opened 2020-06-17
Put Python dependencies (commonmark) in requirements.txt

foolip:


annevk:
Where is this documented?

foolip:
You mean the use of requirements.txt to track Python dependencies? It can be called other things, but https://pip.pypa.io/en/stable/reference/pip_install/#example-requirements-file documents the format.

annevk:
I would prefer something like python-pip.txt or pip.txt to make it more self-describing.

foolip:
requirements.txt is the de-facto standard name for these files, see https://github.com/search?l=&q=filename%3Arequirements.txt&type=Code.

I'm not sure that Dependabot will be able to send PRs if it's named differently, although probably *requirements*.txt work.

Looking into this I found https://github.com/pypa/pipfile which suggests replacing requirements.txt with a Pipfile, which is a more obvious name, but we don't use pipenv. https://github.com/pypa/pipfile#pip-integration-eventual looks promising, but since commonmark doesn't have any dependencies it seems like a complication.

PR #334, opened 2020-06-17
Add a repository_dispatch hook for whatwg/sg updates

foolip:
This is half of the solution, in addition the sg repository will need to
trigger the event:
https://developer.github.com/v3/repos/#create-a-repository-dispatch-event

Part of https://github.com/whatwg/sg/issues/110.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -6,6 +6,10 @@ on:
   push:
     branches:
     - master
+  # Trigger for changes in the whatwg/sg repo.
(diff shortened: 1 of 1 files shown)


foolip:
I'm not 100% confident in `types`, but I based it on https://help.github.com/en/actions/reference/events-that-trigger-workflows#external-events-repository_dispatch and when this has landed I can experiment easily. (Don't feel like creating two other repos for testing.)

PR #335, opened 2020-06-17
Actually deploy for the repository_dispatch event

foolip:
Also change the event name to be a bit more explicit, since the event
name shows up in the list of triggered actions and a longer name will be
easier to spot or search for:
https://github.com/whatwg/whatwg.org/actions

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -9,7 +9,7 @@ on:
   # Trigger for changes in the whatwg/sg repo.
   repository_dispatch:
     types:
-    - sg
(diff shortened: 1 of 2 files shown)


foolip:
The trigger itself worked, https://github.com/whatwg/sg/runs/780782726 triggered https://github.com/whatwg/whatwg.org/actions/runs/138525705, but no actual deploy.

Aside: all of this checking inside of our build scripts is a bit tricky to maintain/understand, and might be improved by splitting the actual deploy into its own step, separate from building, or perhaps having a `--dry-run` argument.

PR #338, opened 2020-06-27
Add stylesheet refinements for MDN annos

sideshowbarker:
This change adds some additional CSS style adjustments related to MDN annos:

* one style adjustment that’s necessary due to the fact we’re sharing styles among both the HTML spec and WHATWG Bikeshed-generated specs.

* one style adjustment that’s necessary due to the fact the existing styles have "position: relative" for domintro boxes, as well as notes, warnings, and examples — so in cases where we have an element with an MDN anno inside of one of those, we need to move it to the right to get it out of containing block and back over to the right margin

### DIFF
--- a/resources.whatwg.org/standard-shared-with-dev.css
+++ b/resources.whatwg.org/standard-shared-with-dev.css
@@ -332,7 +332,9 @@ c-[il] { color: #000000 } /* Literal.Number.Integer.Long */
 .name-slug-mismatch {
   color: red;
 }
-p + .mdn-anno { margin-top: -45px; }
+/* We need this margin adjustment for MDN annos in the HTML spec, but we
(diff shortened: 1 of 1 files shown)


annevk:
Why did this run two checks though? Hmm.

sideshowbarker:
> Why did this run two checks though? Hmm.

Yeah, and one check failed — without any details about the failure, and no log at all

sideshowbarker:
> Yeah, and one check failed — without any details about the failure, and no log at all

OK, I just pushed the button to re-run the second one and now we’re all green (though I still don’t know why it did two checks rather than just one)

annevk:
Feel free to land, if it happens again we can investigate.

PR #342, opened 2020-09-01
Make deploy.sh grammar check not trip over a name

annevk:


annevk:
The motivation here is landing https://github.com/whatwg/encoding/pull/226. I tested locally that it works for that standard.

zcorpan (approved):

PR #343, opened 2020-09-25
Add Bikeshed's style-dfn-panels styles to standard.css

andreubotella:
This change additionally modifies a few rules from Bikeshed's styles for integration with the dfn panels in the HTML spec.

See tabatkins/bikeshed#1772, which turns off Bikeshed's `style-dfn-panels` styles, and whatwg/html#5947, which modifies HTML's `html-dfn.js` script to work with these styles.

### DIFF
--- a/resources.whatwg.org/standard.css
+++ b/resources.whatwg.org/standard.css
@@ -571,3 +571,52 @@ pre > code.css, pre.highlight.lang-css {
 .wpt-tests-margin.wrapped > :not(input) {
   display: none;
 }
+
+/* Dfn panels */
+.dfn-panel {
+    display: inline-block;
+    position: absolute;
+    z-index: 35;
+    height: auto;
+    width: -webkit-fit-content;
+    width: fit-content;
+    max-width: 300px;
+    max-height: 500px;
+    overflow: auto;
+    padding: 0.5em 0.75em;
+    font: small Helvetica Neue, sans-serif, Droid Sans Fallback;
+    background: #ddd;
+    color: black;
+    border: outset 0.2em;
+    cursor: initial;
+}
+.dfn-panel:not(.on) { display: none; }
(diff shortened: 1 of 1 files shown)


annevk:
Thanks for working on this. Can we merge this without merging the HTML change or will HTML be badly impacted if we do that?

andreubotella:
> Thanks for working on this. Can we merge this without merging the HTML change or will HTML be badly impacted if we do that?

Merging this before the HTML change is fine, since the HTML panels use different classes than Bikeshed's (Bikeshed's `dfn-panel` is HTML's `dfnPanel`, and HTML has no equivalent of Bikeshed's `dfn-paneled`).

However, since @sideshowbarker has raised concerns in whatwg/html#5947 about the way I've changed the way the dfn panels are positioned in the HTML spec, it's probably best to wait until that's resolved, so the styles can be adjusted.

resources.whatwg.org/standard.css
```diff
@@ -571,3 +571,52 @@ pre > code.css, pre.highlight.lang-css {
...
+    margin: 0 auto;
+    max-width: calc(100vw - 1.5em - .4em - .5em);
+    max-height: 30vh;
+}
+
+/* Elements that trigger a dfn panel when clicked.
+   The dfn-paneled class is specific to Bikeshed, while dfnEnabled
+   identifies builds of the HTML spec that support dfn panels. */
+.dfn-paneled,
+body.dfnEnabled dfn,
```
domenic, line 615: I guess we probably can't use `:is()` yet, sigh.

domenic (approved):
LGTM, and seems to address @sideshowbarker's comments earlier. Happy to merge as soon as you confirm that it's all good to merge.

sideshowbarker (approved):

PR #344, opened 2020-09-30
review.sh: update LATESTRD field

annevk:
Helps with https://github.com/whatwg/sg/issues/124.

### DIFF
--- a/resources.whatwg.org/build/review.sh
+++ b/resources.whatwg.org/build/review.sh
@@ -19,17 +19,28 @@ header "Creating a git branch with a Review Draft:"
 git checkout master
 git pull
 git checkout -b "review-draft-$(date +'%F')"
+echo ""
+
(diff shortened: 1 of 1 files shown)


domenic (approved):
Looks good, and shellcheck passes. You've tested locally?

annevk:
Yeah, the one thing it does not handle is going from not having this to needing this. I think doing a bit of additional work at that point is fine though.

annevk:
I just realized that one thing that will happen here is that LATESTRD gets copied over to the Review Draft as well. So we have to make a choice:

1. Allow that, but use one of the new Bikeshed features to exclude it from being used.
2. Remove it here as part of the script.

I think I prefer 1 as having less differences between the source documents seems good.

domenic:
I agree with your reasoning.

PR #351, opened 2021-01-26
List Twitter accounts on spec.whatwg.org

### LINKED ISSUE #350: List Twitter accounts on spec.whatwg.org
Once https://github.com/whatwg/meta/issues/109 is fixed (or simultaneously) we should start listing them there to make them easier to find.

annevk:
Closes #350.


domenic (approved):
Hmm, I'm not sure what I think of this being part of the `<dt>` (that seems too "high profile"?) but I don't have any better ideas.

PR #352, opened 2021-02-10
Add websockets logo

domenic:
Part of https://github.com/whatwg/meta/issues/202. @ricea


annevk (commented):
I wonder if we need to say more given https://www.w3.org/html/logo/faq.html#attribute-to and https://creativecommons.org/licenses/by/3.0/. I suspect we need to indicate what changes were made somewhere and maybe we need to use their suggested attribution text?

domenic:
I dunno, I based this on what we do in the acknowledgments sections of standards today.

annevk (approved):

PR #357, opened 2021-03-19
Improve PNG logos

annevk:
This is the result of running `inkscape --export-type=png *.svg -w 500 -h 500` and using https://imageoptim.com/.

inkscape cannot deal with logo-infra.svg so I left that alone.

(One thing this does is fill in the white background for the main logo, but somehow GitHub ignores the transparent parts and uses white there too.)


annevk:
I'm not sure why logo-fetch increased in size, but I suppose it's because it now looks crisper? And maybe logo-quirks isn't needed because it's the same as the normal logo?

domenic (approved):
Seems good, although maybe this should be done by a build process one day.

PR #361, opened 2021-05-12
build.sh: adjust HTML checker download location

annevk:
See this monologue for context: https://freenode.logbot.info/whatwg/20210512#c7977887.


annevk:
I ran `grep -r "https://github.com/validator/validator/releases/download/linux/vnu.linux.zip" *` after the lol commit and got no hits, but @sideshowbarker might know about other places?

Note that this is actively blocking Fetch https://github.com/whatwg/fetch/commits and any change that is going to be made after that, including any PR.

domenic (approved):
Bizarre.

PR #363, opened 2021-05-27
Add Matrix info; change page title from IRC to Chat

sideshowbarker:
As part of this, I can also set up a redirect on the nginx server, to redirect https://whatwg.org/irc to https://whatwg.org/chat

cc @jgraham for review


sideshowbarker:
#364 is an an alternative to this that adds a separate Matrix page, and adds Matrix info/link directly on the index page, as a sibling to the IRC info/link.

whatwg.org/chat
```diff
@@ -27,38 +27,42 @@
...
 
-<p>Note that if you go to IRC to ask a question, it might take a while to get a reply. It can pay off to stick around for a couple of hours or more.</p>
+<p>Note that if you ask a question in either of those places, it might take a while to get a reply. It can pay off to stick around for a couple of hours or more.</p>
 
-<p>If you want to run a bot, let us know. If they are useful, e.g. providing logging facilities, then they are more than welcome.</p>
+<h3 id="getting-started">Getting Started with IRC or Matrix<a class="self-link" href="#getting-started"></a></h3>
+
+<p>The simplest way to get started with IRC, if you are not familiar, is by signing up for a free <a href="https://www.irccloud.com/">IRCCloud</a> account. Once you've done that, you should be logged in to the Freenode server by default. All you'll have to do is join the #whatwg channel. If you tell your browser to use IRCCloud for irc:// links, then just <a href="ircs://irc.freenode.net/whatwg">clicking this link</a> should also take you there. In order to join the channel, you will need to <a href="https://freenode.net/kb/answer/registration">register your nick on Freenode</a>.</p>
```
jgraham, line 38: I think the positioning of this implies that the simplest way to get started _is_ using IRC, but I don't think that's true; I'd prefer we direct new users to the service that doesn't require magic incantations to avoid being spammed etc.

Generally if the plan is to have both FreeNode and Matrix on an ongoing basis, I'd suggest that we divide this section up into two subsections for the different options (I'd also prefer we just make a decision instead of having two options, but I see why you chose this route for the PR).

whatwg.org/chat
```diff
@@ -27,38 +27,42 @@
...
 
 <h3 id="logs">Logs<a class="self-link" href="#logs"></a></h3>
 
-<p>Logs for the #whatwg channel can be found here:</p>
+<p>Chat logs can be found in the following places:</p>
 
 <ul>
- <li><a href="https://freenode.logbot.info/whatwg">https://freenode.logbot.info/whatwg</a></li>
- <li><a href="https://krijnhoetmer.nl/irc-logs/">https://krijnhoetmer.nl/irc-logs/</a> (not updated since 2016-04)</li>
+ <li><a href="https://freenode.logbot.info/whatwg">https://freenode.logbot.info/whatwg</a> (IRC channel)</li>
```
jgraham, line 49: This is going away in 6 weeks, so it hardly seems worth linking to it.

whatwg.org/chat
```diff
@@ -27,38 +27,42 @@
...
 
-<h3 id="getting-started-with-irc">Getting Started with IRC<a class="self-link" href="#getting-started-with-irc"></a></h3>
-
-<p>The simplest way to get started with IRC, if you are not familiar, is by signing up for a free <a href="https://www.irccloud.com/">IRCCloud</a> account. Once you've done that, you should be logged in to the Freenode server by default. All you'll have to do is join the #whatwg channel. If you tell your browser to use IRCCloud for irc:// links, then just <a href="ircs://irc.freenode.net/whatwg">clicking this link</a> should also take you there. In order to join the channel, you will need to <a href="https://freenode.net/kb/answer/registration">register your nick on Freenode</a>.</p>
-
-<p class="note">In view of ongoing spam messaging on Freenode, we highly suggest you <a href="https://freenode.net/kb/answer/usermodes">set your mode</a> to <code>+r</code> in order to block unregistered users from sending you messages.</p>
+<p>If you want to run a bot, let us know. If they are useful, e.g. providing logging facilities, then they are more than welcome.</p>
```
annevk, line 60:
```suggestion
<p>If you want to run a bot, let us know. If they are useful, e.g., providing logging facilities, then they are more than welcome.</p>
```

whatwg.org/chat
annevk: Let's see in 6 weeks. Ideally we'd link to a replacement or a backup before removing this entirely.

annevk (commented):
I was thinking of calling this Communication, but perhaps Chat is indeed better.

I agree that we should link to Matrix primarily. I'm okay with linking to Freenode as alternative in its own subsection for the time being, but given the conversation in whatwg/meta I would not expect that to be long-lived. We'll see.

whatwg.org/chat
sideshowbarker: > I think the positioning of this implies that the simplest way to get started _is_ using IRC, but I don't think that's true; I'd prefer we direct new users to the service that doesn't require magic incantations to avoid being spammed etc.

OK, ec88c16 is an attempt to reposition things along those lines.

> Generally if the plan is to have both FreeNode and Matrix on an ongoing basis, I'd suggest that we divide this section up into two subsections for the different options

OK, I’ll put together another commit which does that

whatwg.org/chat
sideshowbarker: >> This is going away in 6 weeks, so it hardly seems worth linking to it.
>
> Let's see in 6 weeks. Ideally we'd link to a replacement or a backup before removing this entirely.

For now, I pushed bb25ab3 to show what dropping it would look like. But if we don’t want to drop it, I can just strip out that commit before merging this.

sideshowbarker:
> I was thinking of calling this Communication, but perhaps Chat is indeed better.

Yeah, “Chat” is shorter and more precise too, I guess

> I agree that we should link to Matrix primarily. I'm okay with linking to Freenode as alternative in its own subsection for the time being, but given the conversation in whatwg/meta I would not expect that to be long-lived. We'll see.

OK, I’ll do a commit that makes it a subsection as @jgraham suggested

whatwg.org/chat
annevk: I'd rather not drop it until we actually host the archive ourselves somewhere.

sideshowbarker: OK, I’ve stripped out the commit which dropped the https://freenode.logbot.info/whatwg info

whatwg.org/chat
```diff
@@ -27,38 +27,44 @@
  <a href="https://participate.whatwg.org/">Participate</a>
 </nav>
 
-<h3 id="irc">IRC<a class="self-link" href="#irc"></a></h3>
+<h3 id="chat">Chat<a class="self-link" href="#irc"></a></h3>
 
-<p>The Freenode IRC network has a channel called <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> where some of the more active WHATWG community members hang out. Feel free to join us!</p>
+<p>We have both a <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> channel on the Freenode IRC network and a <a href="https://app.element.io/#/room/#whatwg:matrix.org">whatwg</a> room in the <a href="https://en.wikipedia.org/wiki/Matrix_(protocol)">Matrix</a> network. Join us!</p>
```
jgraham, line 32: I'd also like to turn this sentence around, something like:
```
<p>We have a  a <a href="https://app.element.io/#/room/#whatwg:matrix.org">whatwg</a> room in the <a href="https://en.wikipedia.org/wiki/Matrix_(protocol)">Matrix</a> network, as well as an older <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> channel on the Freenode IRC network. Join us!</p>

Comment (sideshowbarker) [reply whatwg.org/chat]: > I'd also like to turn this sentence around, something like:
> 
> ```
> <p>We have a  a <a href="https://app.element.io/#/room/#whatwg:matrix.org">whatwg</a> room in the <a href="https://en.wikipedia.org/wiki/Matrix_(protocol)">Matrix</a> network, as well as an older <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> channel on the Freenode IRC network. Join us!</p>
> ```

OK — made it so

File: whatwg.org/index.html [hunk 5]
```diff
@@ -169,8 +169,8 @@ <h2>Maintaining and evolving HTML since 2004</h2>
   <strong>Participate</strong>
   <p>Get started with contributing to the WHATWG</p>
  </a>
- <a href="/irc" id="irc">
-  <strong>IRC</strong>
+ <a href="/chat" id="chat">
+  <strong>Chat</strong>
   <p>Chat with other members of the WHATWG community</p>
```
annevk, line 174:
```suggestion
  <p>Talk to other members of the WHATWG community</p>
```

annevk (approved):
This looks good to me as an initial take.

One thing we should also do is remind people of the Code of Conduct here, but that can be a follow-up.

jgraham (approved):

sideshowbarker:
> One thing we should also do is remind people of the Code of Conduct here, but that can be a follow-up.

I will get that added right now

whatwg.org/chat
```diff
@@ -27,38 +27,46 @@
...
 
-<p>The Freenode IRC network has a channel called <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> where some of the more active WHATWG community members hang out. Feel free to join us!</p>
+<p>We have a <a href="https://app.element.io/#/room/#whatwg:matrix.org">WHATWG</a> room in the <a href="https://en.wikipedia.org/wiki/Matrix_(protocol)">Matrix</a> network, and an older <a href="ircs://irc.freenode.net/whatwg">#whatwg</a> channel on the Freenode IRC network. Join us!</p>
 
-<p>Note that if you go to IRC to ask a question, it might take a while to get a reply. It can pay off to stick around for a couple of hours or more.</p>
+<p>Note that if you ask a question in either of those places, it might take a while to get a reply. It can pay off to stick around for a couple of hours or more.</p>
 
-<p>If you want to run a bot, let us know. If they are useful, e.g. providing logging facilities, then they are more than welcome.</p>
+<p>We are committed to providing a friendly, safe, and welcoming environment for all. So please read the <a href="https://whatwg.org/code-of-conduct">WHATWG Code of Conduct</a> and respect it when participating in discussions.</p>
```
annevk, line 36:
```suggestion
<p>We are committed to providing a friendly, safe, and welcoming environment for all. So please read the <a href="/code-of-conduct">WHATWG Code of Conduct</a> and respect it when participating in discussions.</p>
```

annevk (approved):

PR #365, opened 2021-05-28
Update remaining /irc links to /chat

annevk:


sideshowbarker (approved):
Sorry for not having caught these myself (especially the necessary SW updates to the filenames).

PR #381, opened 2021-10-10
docs: add “breadcrumb” per suggestion from @annevk

j9t:
Cf. https://github.com/whatwg/html/pull/7165#pullrequestreview-774691072

Signed-off-by: Jens Oliver Meiert <jens@meiert.com>


annevk:
@j9t thanks!

Minor request: please don't at-mention folks in commit messages as they'll get a lot of additional notifications they don't need.

PR #385, opened 2021-12-01
Sort spec.whatwg.org by name of standard

### LINKED ISSUE #384: convert_sg_db.py should alphabetize by standard name
When I added WebSockets to the HTML Workstream we end up with this:

image

IMO the spec.whatwg.org page should just contain the standards, not grouped by workstream. So it should instead put WebSockets down with Web IDL.

annevk:
Otherwise we end up leaking Workstreams here, which is a little ugly.

Fixes #384.


annevk:
@jgraham perhaps you're interested in reviewing this? I suspect there's also some Python tricks to make this shorter, though on the other hand this is quite readable (to me).

domenic (approved):
LGTM, confirmed working

annevk:
Happy to refactor if folks have suggestions.

PR #396, opened 2022-04-29
Add support for <dl class="props">

domenic:
This makes declaring struct instances (in the sense of https://infra.spec.whatwg.org/#structs) much easier.

---

Extracted from https://github.com/whatwg/html/pull/6315. See some example usages at https://whatpr.org/html/6315/history.html#snapshotting-source-snapshot-params and https://whatpr.org/html/6315/history.html#navigate steps 14.7 and 14.13.6.

### DIFF
--- a/resources.whatwg.org/standard-service-worker.js
+++ b/resources.whatwg.org/standard-service-worker.js
@@ -8,7 +8,7 @@
 
 const standardShortname = location.host.split(".")[0];
 
-const cacheKey = "v7";
+const cacheKey = "v8";
 const toCache = [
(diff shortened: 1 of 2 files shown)


annevk:
@tabatkins was asking for a struct literal syntax on Matrix. We should probably pick one or the other?

domenic:
I think this is useful even when you don't want a literal, e.g. the examples I linked where the structs are large and so you specifically want a column-based presentation.

annevk (approved):
If we add literal syntax let's try to remember to add a contrasting example then.

PR #402, opened 2022-10-08
Add Steering Group logo

annevk:
This will probably not be used much, if at all, but I liked the idea of having one and hober had a great suggestion.

---

The SVG could potentially be further optimized by removing the part of the path that makes the outer circle and using stroking for that instead. That seemed like more effort than I was willing to put into it however.

This is based on https://www.svgrepo.com/svg/232240/steering-wheel-car (CC0).


domenic (approved):
Fun!

cwilso (approved):

tantek (approved):

PR #403, opened 2022-10-08
Compress several logos a bit more

annevk:
ImageOptim FTW.


domenic (approved):
Unlike previous attempts, this time I don't see color differences!

PR #404, opened 2022-10-13
Use locally-installed Bikeshed instead of CSSWG API

domenic:
The CSSWG API is often down, and it's bad for it to block our builds and deploys. The pypi package repository is presumably down less.

This should be rolled out only after https://github.com/whatwg/spec-factory/pull/42 is merged and deployed across repos.


domenic:
I guess this was opposed in https://github.com/whatwg/spec-factory/pull/42 and hasn't been a huge problem recently, so closing.

Issue #401, opened 2022-09-15
Sort out hgroup usage


annevk:
We need to replace `h2` with `p` and adjust stylesheets while at the same time not upsetting the validator.

cc @sideshowbarker

domenic:
PR list:

- [ ] whatwg.org, including the stylesheets themselves: https://github.com/whatwg/whatwg.org/pull/408
- [ ] Bikeshed and thus all non-HTML WHATWG specs: https://github.com/tabatkins/bikeshed-boilerplate/pull/29#pullrequestreview-1168028104
- [ ] HTML: https://github.com/whatwg/html/pull/8473

domenic:
All PRs merged, although Bikeshed specs are not yet regenerated.

PR #408, opened 2022-11-03
Styling hgroup p

annevk:
Needed for https://github.com/whatwg/whatwg.org/issues/401.

---

@domenic this sorta does the thing, but I haven't tested it and it could certainly be organized better. Though some of this was already a mess with sometimes using `.head` and sometimes `hgroup` (as far as I can tell that comes down to the same thing).

Probably also want to look at the style inspector to make sure there's not a lot of rules getting dropped and such.

### DIFF
--- a/resources.whatwg.org/standard.css
+++ b/resources.whatwg.org/standard.css
@@ -54,11 +54,11 @@ code :link, :link code, code :visited, :visited code { color: #CE3C05; }
 pre :link, pre :visited { color: inherit; }
 
 html, ::before { font: 1em/1.45 Helvetica Neue, sans-serif, Droid Sans Fallback; }
-h1, h2, h3, h4, h5, h6 { text-align: left; text-rendering: optimizeLegibility; }
-h1, h2, h3 { color: #3c790a; background: transparent; }
+h1, h2, h3, h4, h5, h6, hgroup p { text-align: left; text-rendering: optimizeLegibility; }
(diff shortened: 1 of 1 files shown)


domenic:
I think this is good now. Indeed it seems like our heading styles are kind of a mess and could be cleaned up a good bit, but, let's leave that for later...

resources.whatwg.org/standard.css
```diff
@@ -368,12 +368,12 @@ td.non-rectangular-cell-indentation { border-top-style: hidden; min-width: 2em;
 .hide { display: none }
 
 .head { margin: 0 0 1em; padding: 1em 0 0 0; display: block; }
-.head p { margin: 0; }
+.head p:not(hgroup p) { margin: 0; }
 .head h1 { margin: 0 100px 0 0; }
-hgroup p { margin: 0 100px 1em 0; }
+hgroup p, .head h2 { margin: 0 100px 1em 0; }
```
annevk, line 373: Does `.head h2` still exist then?

annevk:
This looks good to me, but @domenic you'll have to approve it due to GitHub not being able to figure out who did the majority of the work...

domenic (approved):

resources.whatwg.org/standard.css
domenic: Hmm, seems unlikely, not sure why I did that...

PR #410, opened 2022-11-23
Meta: bump actions/setup-python

annevk:
As you can see in this log `@v3` issues a warning: https://github.com/whatwg/whatwg.org/actions/runs/3410148383/jobs/5672752326. Let's see if that warning disappears.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -21,7 +21,7 @@ jobs:
         python-version: "3.10"
     - uses: actions/setup-node@v3
       with:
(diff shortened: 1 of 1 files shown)


annevk:
No warning. Bit annoying to have to bump everything again, but we have until June next year.

domenic (approved):

annevk:
Done, now I wonder if we should theoretically quote that as well as we do with Python version. I guess it doesn't matter much if it stays an integer.

PR #411, opened 2022-11-30
bugfix: generate section with id

endersonmenezes:
Other pages used hyperlinks to this one with anchors links. We want to fix the following page: https://wicg.github.io/admin/charter.html#liaisons which has links to this page, but it would be easier if the IDs were working, we believe that just inserting this id in the code will help a lot.

Co-authored-by: Guilherme Siquinelli <guiseek@gmail.com>

### DIFF
--- a/convert_sg_db.py
+++ b/convert_sg_db.py
@@ -85,7 +85,7 @@ def create_workstreams(db, template):
     introduction = introduction.replace("<!-- This ends up being included at https://whatwg.org/workstreams -->\n", "")
     content = markdown(introduction, obtain_link_mapping())
     for workstream in db["workstreams"]:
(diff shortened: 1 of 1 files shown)


domenic (approved):

endersonmenezes:
@domenic Is there any other page that interferes with the build of this one? Why did the ID's disappear from the site, right after the merge they appeared there and I checked.

annevk:
I suspect you might be seeing the service worker, though that too should update after a refresh. When I test here in Private Browsing it seems to work fine.

endersonmenezes:
It's all working, we had the other PR also approved! Thank you very much and sorry for the beginner doubts.

PR #418, opened 2023-05-15
Recommend loading commit-snapshot-shortcut-key.js with defer as it installs a DOMContentLoaded listener

jsnajdr:
Fixing what I believe is a little bug in the `commit-snapshot-shortcut-key.js` script. In `html.spec.whatwg.org` it's loaded with the `async` flag, which means it can run after the `DOMContentLoaded` event has fired. The listener that sets up the keyboard shortcut will never run in that case. Only a script with `defer` is guaranteed to run before `DOMContentLoaded`.

The fix uses the `load` event instead, which is fired only after even the `async` scripts are finished running.

In the spec, [this is the place](https://html.spec.whatwg.org/#the-end:set-of-scripts-that-will-execute-as-soon-as-possible) where all the `async` scripts are flushed, and it runs between the `DOMContentLoaded` and `load` events.

I found this when trying to learn what exactly the difference between `async` and `defer` is.

### DIFF
--- a/resources.whatwg.org/commit-snapshot-shortcut-key.js
+++ b/resources.whatwg.org/commit-snapshot-shortcut-key.js
@@ -2,7 +2,7 @@
 // `<a href="https://{my-spec}.spec.whatwg.org/commit-snapshots/{commit-sha}" id="commit-snapshot-link">Snapshot as of this commit</a>`.
 //
 // Then include this script with
-// `<script src="https://resources.whatwg.org/commit-snapshot-shortcut-key.js" async></script>`.
+// `<script src="https://resources.whatwg.org/commit-snapshot-shortcut-key.js" defer></script>`.
 
 (function () {
   'use strict';


annevk:
Hmm, perhaps we should be using `defer` instead? We don't really want to wait on everything else being loaded either after all.

jsnajdr:
In this case, where all the code is in the `DOMContentLoaded`/`load` listener, `defer` wouldn't make much difference. `defer` script are guaranteed to be executed only after "the document has finished parsing", immediately before `DOMContentLoaded`. We'd be waiting for everything else to be loaded anyway.

Also, at the time when the listener runs, the `a#commit-snapshot-link` element must already be present in the document.

So, yes, both combinations, `DOMContentLoaded` + `defer`, and `load` + `async`, are correct (but not the `DOMContentLoaded` + `async` one!), but they do almost exactly the same thing.

annevk:
That's not entirely true. We don't have to wait for things that delay the load event, such as images.

jsnajdr:
> We don't have to wait for things that delay the load event, such as images.

Oh, this is a good point -- I thought that the `DOMContentLoaded` and `load` events are very close together, but they aren't. We're waiting until all the resources that are neither parser-blocking nor render-blocking get loaded.

I'll rework both patches to use `defer` instead, and keep the `DOMContentLoaded` listeners.

jsnajdr:
This PR has been transformed into one that merely changes the comment that describes how the script should be included. All other work is now done in https://github.com/whatwg/html/pull/9297.

annevk:
For the non-HTML standards https://github.com/speced/bikeshed-boilerplate/blob/main/boilerplate/whatwg/header.include will have to be patched. Do you want to take that on as well?

jsnajdr:
> Do you want to take that on as well?

Yes, done in https://github.com/speced/bikeshed-boilerplate/pull/40.

annevk (approved):

annevk:
Thanks for following through on everything here @jsnajdr! Each specification will need a new commit before this change becomes part of it, but hopefully that doesn't take too long overall.

PR #423, opened 2023-09-07
Fix a few issues in standard.css

yarusome:
This PR fixes a few visible issues in the style sheet. Please see comments below for explanations.

### DIFF
--- a/resources.whatwg.org/standard.css
+++ b/resources.whatwg.org/standard.css
@@ -261,14 +261,23 @@ span.note::before, span.warning::before {
 .domintro dt, .domintro dt * {
   color: black;
 }
-dl.domintro dd p {
+.domintro dd p {
   margin: 0.5em 0;
 }
-dl.domintro dd p:first-child {
+.domintro dd p:first-child {
   margin-top: 0;
 }
-dl.domintro dd p:last-child {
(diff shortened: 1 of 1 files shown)


resources.whatwg.org/standard.css
```diff
@@ -304,6 +313,9 @@ dl.domintro dd p:last-child {
 .example::before {
   background: #222222;
 }
+td > .example, td > .note:not(span), td > .warning:not(span) {
+  margin-left: 0.8em;
+}
 td > .example:only-child {
   margin: 0 0 0 0.1em;
   padding: 0;
```
yarusome, line 321: This fixes [13.2.2 Parse errors](https://html.spec.whatwg.org/multipage/parsing.html#parse-error-invalid-first-character-of-tag-name), where `.example::before` and `.note::before` intrude into adjacent cells.

annevk:
Thank you, these improvements seem great. @domenic do you want to have a look as well?

domenic:
Maybe add comments in the CSS, in addition to notes in a self-review of the PR?

yarusome:
@domenic I just added some comments. What do you think?

domenic (approved):
Thanks!

PR #424, opened 2023-09-26
Avoid generating review drafts for merge commits

dbaron:
This avoids generating review drafts for merge commits unless the merge  commit has modifications to the review draft (which I hope means that it's different from all parents of the merge).

The code prior to this change tries to generate a review draft if the  first parent of the merge commit introduces a review draft. (This happens if "git merge main" is done on a feature branch and main has introduced a review draft.)

This matches the change in 
https://github.com/whatwg/html-build/pull/284 .

### DIFF
--- a/resources.whatwg.org/build/deploy.sh
+++ b/resources.whatwg.org/build/deploy.sh
@@ -122,7 +122,7 @@ echo ""
 
 header "Starting review drafts (if applicable)..."
 echo "Note: review drafts must be added or changed in a single commit on main"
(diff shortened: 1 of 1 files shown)


annevk (approved):
Interesting, we have a policy to never do merge commits on main, but I guess this is still useful for pull requests where no such thing is enforced.

Rubber-stamp based on it being the same as the other change.

dbaron:
Indeed, it's for pull requests... and I do think merge commits are useful on pull requests since they let the commit history of the pull request reflect reality.

PR #434, opened 2024-02-07
Darkmode the remaining colors used in a few HTML spots.

tabatkins:
Sorry, missed several specialized bits on the first pass, because I was just looking at DOM/etc, not HTML. Every single `color`, `background`, and `border` declaration in `standard.css` has been double-checked now.

* The `box-shadow` colors on the first-page nav blocks in HTML are kept as-is rather than being abstracted; once I swapped out the transparent overlay from nearly-white to nearly-black, the specified colors ended up still looking great, imo.
* I think that the "file a bug from selected text" functionality was removed, so possibly we can just kill that style.
* I also couldn't find any `tr.rare` elements, so I had to just guess at a reasonable color swap. I verified they looked reasonable on some random text, but I'm not sure if there's some particulars in context that would make it weird. Probably not, tho.

### DIFF
--- a/resources.whatwg.org/standard-service-worker.js
+++ b/resources.whatwg.org/standard-service-worker.js
@@ -8,7 +8,7 @@
 
 const standardShortname = location.host.split(".")[0];
 
-const cacheKey = "v10";
+const cacheKey = "v11";
 const toCache = [
   location.origin + "/",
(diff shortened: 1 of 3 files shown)


domenic (approved):
Thank you so much!

PR #443, opened 2024-05-18
Move twitter.com hrefs over to x.com

TylerJMorg:
As of 2024-05-17, X (formerly Twitter) has migrated their hosting domain over to x.com and twitter.com now pulls a HTTP 302 redirect whereas x.com is 200.

Renaming the "twitter" values in db.json and db.py in the sg repo, and in convert_sg_db.py (and elsewhere) is up for discussion if it's truly needed.


domenic (approved):

domenic:
Although, we should probably update various other instances, e.g. https://whatwg.org/ itself.

TylerJMorg:
I made that change, I couldn't any other instances on whatwg/whatwg.org or whatwg/sg. Although should we switch whatwg.org/whatwg.org/img/bird-webpage.svg to an "X" SVG?

domenic:
Yes, that would be nice. Thanks for doing that. We should also delete the old one. Let me give it a shot...

PR #444, opened 2024-05-21
Support dark mode in tracking-vector.svg

zcorpan:
See https://github.com/w3c/tr-design/issues/162#issuecomment-2122073793

Demo: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/12744

This seems to work in Gecko and Chromium, but not WebKit.


zcorpan:
https://bugs.webkit.org/show_bug.cgi?id=199134

domenic (approved):

PR #445, opened 2024-05-24
Create logo-compression.svg

saschanaz:
Corresponding to https://github.com/whatwg/sg/issues/231.

I have no idea how SVG works, but this at least works with no compat issue between Firefox/Chrome/Safari and Inkscape.

Originally I tried what Anne suggested https://github.com/whatwg/sg/issues/231#issuecomment-2129595476, but then I found other SVGs are greatly minimal so I joined the challenge. 🙂

I put a line between the arrows though as otherwise it just looks like collision.


annevk (approved):
This looks great!

PR #448, opened 2024-07-13
Add phonetic pronunciation for WHATWG

tylerjmorg:
I figured that it may be useful to add IPA pronunciations for the different examples on the FAQ page.

Small change, but may be useful for internationalization of the name. :)

### DIFF
--- a/whatwg.org/faq
+++ b/whatwg.org/faq
@@ -50,22 +50,19 @@ governance structure for the WHATWG, together forming a
 
 <h3 id="spell-and-pronounce">How do you spell and pronounce WHATWG?<a class="self-link" href="#spell-and-pronounce"></a></h3>
 
-<p>It is spelled WHATWG, all uppercase, no spaces. 
-
-<p>It has various pronunciations:
-
+<p>It is spelled WHATWG, all uppercase, no spaces. It has various pronunciations:
 <dl>
  <dt>what-wee-gee
- <dd>IPA US: /wɑːt.wi.dʒi/
- <dd>IPA UK: /wɒt.wi.dʒi/
+ <dd>US: /wɑːt.wi.dʒi/
+ <dd>UK: /wɒt.wi.dʒi/
 
  <dt>what-wig
- <dd>IPA US: /wɑːt.wɪɡ/
- <dd>IPA UK: /wɒt.wɪɡ/
+ <dd>US: /wɑːt.wɪɡ/
+ <dd>UK: /wɒt.wɪɡ/
 
  <dt>what-double-you-gee
- <dd>IPA US: /wɑːt.ˈdʌb.əl.juː.dʒi/
- <dd>IPA UK: /wɒt.ˈdʌb.əl.juː.dʒi/
+ <dd>US: /wɑːt.ˈdʌb.əl.juː.dʒi/
+ <dd>UK: /wɒt.ˈdʌb.əl.juː.dʒi/
 </dl>
 
 <h3 id="what-is-the-whatwg-working-on">What is the WHATWG working on?<a class="self-link" href="#what-is-the-whatwg-working-on"></a></h3>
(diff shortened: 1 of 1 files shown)


whatwg.org/faq
```diff
@@ -50,8 +50,20 @@ governance structure for the WHATWG, together forming a
 
 <h3 id="spell-and-pronounce">How do you spell and pronounce WHATWG?<a class="self-link" href="#spell-and-pronounce"></a></h3>
 
-<p>It is spelled WHATWG, all uppercase, no spaces. It has various pronunciations: what-wee-gee,
-what-wig, what-double-you-gee.
+<p>It is spelled WHATWG, all uppercase, no spaces. It has various pronunciations:
+<dl>
+ <dt>what-wee-gee
+ <dd>US: /wɑːt.wi.dʒi/
```
annevk, line 56: Instead of US use en-US or American English here and below. Similar for UK.

annevk (commented):
This is pretty neat. Happy to trust you these are correct.

tylerjmorg:
Thank you! That should be fixed for IETF langs.

tylerjmorg:
Also, would you rather en-UK be switched to en-GB?

annevk:
Yeah, en-GB would be better.

There was this feedback on Matrix about this PR:

> They have "what" as rhyming with "lot" in both the UK and US pronunciations – I think at least in the US it's a lot more common to rhyme it with "cut"

whatwg.org/faq
```diff
@@ -50,8 +50,20 @@ governance structure for the WHATWG, together forming a
 
 <h3 id="spell-and-pronounce">How do you spell and pronounce WHATWG?<a class="self-link" href="#spell-and-pronounce"></a></h3>
 
-<p>It is spelled WHATWG, all uppercase, no spaces. It has various pronunciations: what-wee-gee,
-what-wig, what-double-you-gee.
+<p>It is spelled WHATWG, all uppercase, no spaces. It has various pronunciations:
+<dl>
+ <dt>what-wee-gee
+ <dd>en-US: /wɑːt.wi.dʒi/
+ <dd>en-UK: /wɒt.wi.dʒi/
```
gsnedders, line 57: This goes for all three, but is there actually a difference in vowel length between en-US/en-GB here? Because I don't believe there is?

gsnedders (commented):
I wonder how much value there actually is in having the en-US/GB distinction here? There's only any notable distinction in the "what", and arguably even then these are different realisation of the same phonemes.

janbrasna:
Actually, is there a preferred way how to call oneselves by the group/members?

If so, it would be also worth updating https://en.wiktionary.org/wiki/WHATWG#Pronunciation to show the preferred choice, or all acceptable forms.

tylerjmorg:
I'll address these questions based on my perceptive.

1. I've based these pronunciations on the Cambridge Dictionary. So I'm unsure whether we should include multiple en-GB pronunciations which differ based on region. See a3.

2. I agree, the vowel length should be changed especially since "WHAT" is grouped with WG phonetically in all three examples.

3. Kind of tying into the first address. Since WHATWG is steered by Apple, Google, Microsoft, and Mozilla. Which the HQs are based in the U.S., we could just consider the en-US IPA as the primary pronunciation and leave other regions like en-GB to decide their own dialect (e.g., Welsh, Cockney, Yorkshire, etc).

4. WHATWG is pronounced in multiple ways (e.g., I pronounce it as WHAT-DOUBLE YOU-GEE myself). So having just one way to pronounce WHATWG may lead to conflicts and unnecessary habit changes. However, the steering group may feel differently and that can be a discussion at a later time. I just think that think this PR wouldn't be the appropriate channel to discuss that big of a change.

Let me know what y'all think! Love to hear from you all! Should we ditch the en-GB section altogether? Do you think having IPA pronunciations may be unnecessary?

annevk:
Let's see:

- I like having the various pronunciations listed.
- Adding IPA seems useful.
- en-US might well be sufficient. All our other documentation is in en-US(-x-Hixie) so might as well stick with that here.
- And yeah, while what-wee-gee is clearly superior, there's intentionally no right answer.

PR #449, opened 2024-07-22
ImageOptimize PNG logos

### LINKED ISSUE #96: PNG logos have the wrong color in macOS Chrome/Safari
See https://github.com/whatwg/whatwg.org/pull/95#issuecomment-332226057. Per https://webkit.org/blog/6682/improving-color-on-the-web/ the problem might be that the images have a color profile? It seems without sRGB would be assumed, which would presumably do the right thing?

janbrasna:
Trying to resolve #96, this squeezes the last bits of metadata from PNG logos where applicable, to make sure there are no color space variations.

(Previewing the changeset in both WebKit and Gecko on macOS, I see both _"before"_ and _"after"_ green as `#3C7A1C` FWIW, so the color profile issue observed may in the end be something just locally, this all seems A-OK.)


annevk (approved):
Nice!

PR #457, opened 2025-03-07
Align on .yesno styling

annevk:
While .applies is specific to HTML (for now), it seems best to include that rule here.


domenic (approved):

domenic:
Wait, we should probably rev https://github.com/whatwg/whatwg.org/blob/57c7d1906108787539437959a93793f9df1b5f1e/resources.whatwg.org/standard-service-worker.js#L11 in this PR to make sure all the service worker caches get updated.

annevk:
Good call, will do!

PR #476, opened 2025-11-05
Fix vnu download link in deploy scripts

MattiasBuelens:
According to [the GitHub docs](https://docs.github.com/en/repositories/releasing-projects-on-github/linking-to-releases#linking-to-the-latest-release), the download link should be `/releases/latest/download/` but we were using `/releases/download/latest/`. The old URL now returns HTTP 404 and [breaks our deploy script](https://github.com/whatwg/compression/pull/79#issuecomment-3490281004).

### DIFF
--- a/deploy.sh
+++ b/deploy.sh
@@ -50,7 +50,7 @@ if [[ "$GITHUB_ACTIONS" == "true" ]]; then
     readarray -d '' TARGETS < <(find whatwg.org -maxdepth 1 -type f ! -name "*.*" ! -name "status-2008-12" -print0)
     TARGETS+=(whatwg.org/news whatwg.org/validator whatwg.org/index.html idea.whatwg.org/index.html spec.whatwg.org/index.html)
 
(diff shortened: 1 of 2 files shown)


MattiasBuelens:
Good news: the HTML checker is working again!
Bad news: [we have validation errors.](https://github.com/whatwg/whatwg.org/actions/runs/19098497609/job/54564392615?pr=476) 😅

annevk:
Yeah except I think those are wrong. But at this point we really need @sideshowbarker to help us out.

zcorpan:
`p` in `hgroup` is allowed in https://validator.w3.org/nu/ but not in https://validator.nu/

zcorpan:
I think the issue is lack of new releases.

- https://github.com/validator/validator/pull/1467 landed Nov 2, 2022.
- The [latest release](https://github.com/validator/validator/releases) is from 30 June 2020.

annevk:
It seems this wasn't needed after all.

PR #507, opened 2026-09-04
Don't set the commit-snapshot-link title with script

zcorpan:
See https://github.com/whatwg/html/pull/12904


zcorpan:
I haven't yet opened a PR for bikeshed to generate a `title` in markup. I can do so if there's agreement that it's better. cc @tabatkins 

Edit: https://github.com/speced/bikeshed/pull/3322

annevk (approved):

tabatkins:
Why not set it via script? It communicates an action you can only perform when scripting is enabled.

tabatkins:
Ah, you also asked if anyone actually *uses* the keyboard shortcut to do that thing. I suspect the answer is "no" and you can just drop the entire feature.

zcorpan:
That's fine with me. Let's leave the script as-is, but stop using it.

domenic:
I use it...

PR #508, opened 2026-09-14
Use ::marker for dl.switch

zcorpan:
This avoids rendering whitespace between the marker and the text.

This also lets whatwg/wattsi#173 drop its special case for `dl.switch > dt`.


zcorpan:
"Before" in the top list, "after" in the bottom list. The selection shows a leading space caused by `inline-block`, gone with the new style. Markers have the same alignment as the old styles.

Screenshot

annevk (approved):

PR #511, opened 2026-09-23
Add empty alt text for dl.switch > dt::marker content

### LINKED ISSUE #260: <dl class="switch"> styling is not optimal for screen reader users
https://github.com/whatwg/whatwg.org/blob/master/resources.whatwg.org/standard.css#L168

A `<dl class="switch">` in specs represents a `switch` statement in programming languages. For stylistic reasons we use an arrow character as generated content before each `dt`.

Example: https://dom.spec.whatwg.org/#dom-node-nodetype

This renders as "rightwards arrow with hook" in screen readers, according to @mcking65, which is annoying and makes it hard to follow.

It would be better if it read as "Case:"

Maybe also the `<dl>` itself could be introduced with "Switch:" (possibly both visually and in screen readers).

So I think there are a few options. Either we change the styling so that the text "Case:" is rendered for everyone. Or, if we don't want to change the visual rendering, we can use a custom font with a [ligature](https://glyphsapp.com/tutorials/ligatures) that turns "Case:" into a “↪" glyph.

It would be nice to make this change for all specs that use this pattern (I'm aware of WHATWG specs and W3C specs), so they are consistent and accessible. cc @fantasai

6TELOIV:
Fixes #260

This PR adds an empty alt text to the marker content for the switch definition list markers. This *should* make screen readers skip announcing "Left Arrow Curving Right" for every list item.

What I've tested:

- Windows Narrator: Pass
- Android TalkBack: Pass
- Linux Orca: Fail -- it still announces "Left Arrow Curving Right"; so not a regression but no improvement

It would be good if other people using JAWS or NVDA tested as well before closing the issue, as they are much more common screen readers.


zcorpan (approved):

zcorpan:
Which browser did you test in each case? Computing the accessible name happens in the browser a11y tree.

This is covered by https://w3c.github.io/aria/accname/#computation-steps 2.6.2.1:

> For ::marker pseudo elements, [User agents](https://infra.spec.whatwg.org/#user-agent) MUST prepend CSS textual content, without a space, to the textual content of the current node if the current node supports ::marker.

But it's not clear to me that "CSS textual content" means to use the alt text if present. There's an issue at https://github.com/w3c/accname/issues/253

It's tested in https://wpt.fyi/results/accname/name/comp_name_from_pseudo_content_marker.tentative.html , Gecko fails the `alt` cases for `::marker`. Filed https://bugzilla.mozilla.org/show_bug.cgi?id=2075119

6TELOIV:
@zcorpan On Linux, I believe I only tested Firefox. I'll try again this evening in Chromium as well.

